PR #62, opened 2016-11-23
Change workshop to event for terminology reasons

bjrne:
This is needed because currently we only have workshops. In the future, we want to have at least two types of events and a workshop is one of them. We need a precise terminology for this, so workshop was changed to event, which will be the generic term for any kind of event.


bjrne:
@corinnaj Ist geändert, kommt das sonst noch irgendwo im Front-End vor? In der Übersetzungsdatei ist es angepasst.

app/views/events/index.html.erb
```diff
@@ -1,4 +1,4 @@
-<%- model_class = Workshop -%>
+<%- model_class = Event -%>
 <div class="page-header">
   <h1><%=t '.title', :default => model_class.model_name.human.pluralize.titleize %></h1>
```
Baschdl: German plural isn't correct. "Veranstaltungs" instead auf "Veranstaltungen".

PR #64, opened 2016-11-24
Applicants Overview for Events (formerly known as Workshops)

timbasel:
This is basic implementation of the Applicants Overview for the Events Page.

## TODO

- [x] Add localization via I18n
- [x] Only allow Admins, Organizers and Coaches to view the applicants list
- [x] Add general information to application details page
- [x] Improve UI

Issue #11, opened 2016-11-10
US_1.1: Consistent german language 🍪


tobiduer:
**As** 
user 

**I want to** 
have consistent german language in the GUI 

**in order to** 
have a great user experience. 

**5 🍪**

This is an important issue for the already existing system. In the future we expect this to be considered throughout the new features. The responsibility is restricted to tabs that no other team actively modifies.

- [ ] There is no mix of languages for names of input fields, error messages and the like

MasterCarl:
As @TimBasel pointed out, localizations should be implemented using i18n (as in pull request #64)

bjrne:
Adressed in more detail in #145 - #151.

PR #71, opened 2016-11-26
46 1.7 Additional information (events)

MasterCarl:


app/views/events/_form.html.erb
```diff
@@ -43,6 +43,20 @@
     </div>
     <%=f.error_span(:active) %>
   </div>
+  <div class="form-group">
+    <%= f.label :veranstalter, :class => 'control-label col-lg-2' %>
```
TimBasel: should be english, to stay consistent with all the other code (+ all other occurences of german words)

MasterCarl: I made this label in German to comply with #11 since it's part of the GUI. Nevertheless, I agree that it's inconsistent with the rest. Any ideas for better localization?

TimBasel: use i18n (see pull request #64) and translate it there

MasterCarl: Resolved in 0138b956064c37ab4b2e0bf7dad2ae03a3ef4ff6

config/locales/de.yml
```diff
@@ -51,3 +51,8 @@ de:
         country: "Land"
         graduates_school_in: "Abitur vorraussichtlich in"
         already_graduated: "Bereits Abitur"
+  activerecord:
```
TimBasel: Labeling look good, but I think it would be better to split the translation into different files (see merged PR #64) to prevent that the de.yml becomes to large and unreadable

MasterCarl: Good idea, the newly created localizations weren't yet merged into dev. I moved them just now

db/migrate/20161119134300_add_organizer_to_workshops.rb
```diff
@@ -0,0 +1,5 @@
+class AddOrganizerToWorkshops < ActiveRecord::Migration
+  def change
+    add_column :workshops, :organizer, :string
```
bjrne: This really breaks stuff, I don't know why @TimBasel didn't test this. A follow-up PR for this is already on its way, i'll make sure it contains all fixes for this.

MasterCarl: Does it really break stuff? This migration comes before 20161123124500_change_workshop_to_event. At that point in migrations, there is no table "events". The column is added to workshops. Afterwards, the table workshops is renamed to events.

MasterCarl: I think it is really out of scope for the corresponding user story to unify migrations, which breaks compatibility with earlier versions and would require a database drop.

bjrne: The problem is, migrations that were merged earlier are already performed, so you cannot just squeeze yours in. Please rename the file and change the migration. Have a look at [this](http://edgeguides.rubyonrails.org/active_record_migrations.html) or go to the slack channel "ruby-help".

MasterCarl: Did you try running `bin/rails db:rollback` and `bin/rails db:migrate` again?

ekrebs5: This migration only leads to problems if you migrate rename first and then this migration. But since rename was on dev first, this one should be changed, since everyone who already migrated that will have a problem. (My database works because I resetted and migrated again, but like bjarne said, this should not be necessary)

MasterCarl: It should now be fixed in https://github.com/hpi-swt2/workshop-portal/pull/80

PR #73, opened 2016-11-27
24 3.4 notes on applications

frederike-ramin:
Adds the ability to add and view notes on applications. (Only for coaches and organizers)


timbasel (changes requested):

TimBasel:
Rest looks good

app/models/ability.rb
```diff
@@ -42,10 +42,11 @@ def initialize(user)
     if user.role? :tutor
```
cmfcmf: > it was always coach..suddenly: tutor..:x

We originally named it `tutor` because that's how it was named in one of our user stories. But `coach` it is I guess?

frederike-ramin: Maybe that's also a lack of communication between teams. Whit us, the word "tutor" was never even mentioned.

config/routes.rb
```diff
@@ -1,6 +1,8 @@
 Rails.application.routes.draw do
   resources :requests
-  resources :application_letters, path: 'applications'
+  resources :application_letters, path: 'applications' do
+    resources :application_notes #TODO: don't allow new, edit etc routes for notes (only create)
```
frederike-ramin: Fixed via 7c2ea5933735319b7a26749049152740f1329b18.

timbasel (approved):

PR #76, opened 2016-11-28
17 1.3 event time spans

tom95:
This branch adds the time spans for events required by US 1.3.

One test is currently commented as it requires js to run. I prepared a setup for testing with javascript in the separate javascript-tests branch, which I would propose for merging after this branch is merged.


app/controllers/events_controller.rb
```diff
@@ -51,6 +59,21 @@ def set_event
...
+    # are limited in the way they can arrange data for us. So we translate
+    # pairs of { start_date: [{day, month, year}, {day, month, year}], end_date: [...] }
+    # to the expected format [{start_date, end_date}, ...]
+    #
+    # @return array of start_date, end_date pairs
+    def date_range_params
+      dateRanges = params[:date_ranges] || {start_date: [], end_date: []}
+
+      dateRanges[:start_date].zip(dateRanges[:end_date]).map do |s, e|
+        DateRange.new(start_date: Date.new(s[:year].to_i, s[:month].to_i, s[:day].to_i),
```
MasterCarl: Date conversion ( `Date.new(...)` )might be refactored into a separate function to improve readability

app/models/date_range.rb
```diff
@@ -0,0 +1,49 @@
...
+
+
+  belongs_to :event
+
+  validate :validate_start_date_not_in_past
+  validate :validate_end_not_before_start
+
+  def validate_start_date_not_in_past
+    if start_date < Date.today
+      errors.add(:start_date, "darf nicht in der Vergangenheit liegen.")
```
MasterCarl: The german language should be refactored by using i18n, but I think this is out of scope for this user story. @bjrne Any thoughts?

app/models/date_range.rb
```diff
@@ -0,0 +1,49 @@
...
+    if args[0].to_s == "start_date"
+      return "Anfangs-Datum"
+    elsif args[0].to_s == "end_date"
+      return "End-Datum"
+    end
+
+    # NOTE: In our quest for 100% code coverage we can't have this line.
+    # If anyone is to add a new attribute that uses the default label,
+    # reenable this line.
+    # super
```
MasterCarl: I don't think this is an appropriate way to reach 100% code coverage

app/models/event.rb
```diff
@@ -6,12 +6,50 @@
...
 #  description      :string
 #  max_participants :integer
+#  date_ranges      :Collection
 #  active           :boolean
 #  created_at       :datetime         not null
 #  updated_at       :datetime         not null
 #
+
 class Event < ActiveRecord::Base
+  UNREASONABLY_LONG_DATE_SPAN = 300
```
MasterCarl: As a configuration parameter, I think this doesn't really belong into the model class

app/views/events/_form.html.erb
```diff
@@ -3,13 +3,18 @@
   <% if @event.errors.any? %>
     <div id="error_expl" class="panel panel-danger">
       <div class="panel-heading">
-        <h3 class="panel-title"><%= pluralize(@event.errors.count, "error") %> prohibited this event from being saved:</h3>
+        <h3 class="panel-title"><%= pluralize(@event.errors.count, "Fehler") %> verhinderte das Speichern:</h3>
```
MasterCarl: Localization should be used in this case

app/views/events/_form.html.erb
```diff
@@ -44,6 +49,34 @@
...
   </div>
   <div class="form-group">
+    <%= f.label :date_ranges, :class => 'control-label col-lg-2' %>
+    <div class="col-lg-10" id="event-date-pickers">
+      <% def date_picker_template(startDate = Date.today, endDate = Date.today)
+          '<div class="event-date-picker"><div class="event-date-picker-start">' + 
+          select_date(startDate, {prefix: 'date_ranges[start_date][]'}).gsub("\n", "") +
+          '</div> bis <div class="event-date-picker-end">' +
+          select_date(endDate, {prefix: 'date_ranges[end_date][]'}).gsub("\n", "") +
+          '</div></div>'
```
MasterCarl: This hunk might just benefit from string interpolation, date calculations should be extracted

app/views/events/new.html.erb
```diff
@@ -1,5 +1,5 @@
 <%- model_class = Event -%>
 <div class="page-header">
-  <h1><%=t '.title', :default => [:'helpers.titles.new', 'New %{model}'], :model => model_class.model_name.human.titleize %></h1>
+  <h1><%=t '.title', :default => [:'helpers.titles.new', 'Neues %{model}'], :model => model_class.model_name.human.titleize %></h1>
```
MasterCarl: See above

spec/controllers/events_controller_spec.rb
```diff
@@ -19,12 +19,35 @@
...
+  let(:valid_attributes_post) {
+    {
+        event: {
+          name: 'Test',
+          description: 'Test',
+          max_participants: 1,
+          active: false,
+        },
+        date_ranges: {
+          start_date: [
```
MasterCarl: You could use a map function for better readability, but for this test it should be acceptable

spec/factories/events.rb
```diff
@@ -15,9 +16,39 @@
...
+        event.date_ranges << FactoryGirl.create(:date_range, start_date: Date.today.next_day(12), end_date: Date.today.next_day(16))
+        event.date_ranges << FactoryGirl.create(:date_range, start_date: Date.today, end_date: Date.tomorrow)
+      end
+    end
+
+    trait :with_unreasonably_long_range do
+      after(:build) do |event|
+        event.date_ranges << FactoryGirl.create(:date_range,
+          start_date: Date.today,
+          end_date: Date.tomorrow.next_day(Event::UNREASONABLY_LONG_DATE_SPAN))
```
MasterCarl: This is tightly coupled to the event model

spec/features/event_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+    #  page.find("#event-add-date-picker").click
+
+    #  start = page.all('.event-date-picker')[1].find('.event-date-picker-start')
+    #  select_date_within_selector(second_from, start)
+    #  select_date_within_selector(second_to, '.event-date-picker:nth-child(2) .event-date-picker-end')
+    #  click_button "Veranstaltung erstellen"
+
+    #  expect(page).to have_text(first_from.to_s + ' bis ' + first_to.to_s)
+    #  expect(page).to have_text(second_from.to_s + ' bis ' + second_to.to_s)
+    #end
```
MasterCarl: This should be part of a later pull request when js support is added

spec/features/event_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+    end
+  end
+
+  describe "show page" do
+    it "should display all date ranges" do
+      event = FactoryGirl.create(:event, :with_two_date_ranges)
+      visit event_path(event.id)
+
+      expect(page).to have_text(event.date_ranges.first.start_date.to_s +
+                                ' bis ' + event.date_ranges.first.end_date.to_s)
```
MasterCarl: The test shouldn't be dependent on the language and wording used. Maybe you could check for both date strings individually?

spec/features/event_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+                                ' bis ' + event.date_ranges.second.end_date.to_s)
+    end
+  end
+
+  describe "edit page" do
+    it "should display all existing date ranges" do
+      event = FactoryGirl.create(:event, :with_two_date_ranges)
+      visit edit_event_path(event.id)
+
+      expect(page.all('.event-date-picker').size).to eq(2)
```
MasterCarl: This could be improved by using the `assert_select` function, but it's okay

spec/features/event_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+      picker = page.all('.event-date-picker')[0]
+      startPicker = picker.find('.event-date-picker-start')
+      endPicker = picker.find('.event-date-picker-end')
+
+      select_date_within_selector(dateStart, startPicker)
+      select_date_within_selector(dateEnd, endPicker)
+
+      click_button 'Veranstaltung aktualisieren'
+
+      expect(page).to have_text(dateStart.to_s + ' bis ' + dateEnd.to_s)
```
MasterCarl: See above

spec/models/event_spec.rb
```diff
@@ -15,8 +15,45 @@
...
+    it "should return return its minimum over all date ranges" do
+      event = FactoryGirl.create :event, :with_multiple_date_ranges
+      expect(event.start_date).to eq(Date.today)
+    end
+  end
+
+  describe "#end_date" do
+    it "should return return its maximum over all date ranges" do
+      event = FactoryGirl.create :event, :with_multiple_date_ranges
+      expect(event.end_date).to eq(Date.today.next_day(16))
```
MasterCarl: Why is this 16 (magic number)?

app/models/date_range.rb
bjrne: I think since i18n wasn't mentioned as a goal for this sprint we can leave it out for now. But somehow everyone from other teams is going crazy to implement it in different ways. I don't want to decide this, but from my point of view german should be okay for this PR.

app/models/date_range.rb
tom95: Do you have suggestions on how to handle this more well? Just accept that we don't have 100%?

spec/features/event_spec.rb
tom95: @MasterCarl I tend to disagree. Feature tests all explicitly focus on what is actually displayed to the user and in this particular case it is of vital importance that it is a date range and as such clear to the user and not dates scattered somewhere on the page.

spec/models/event_spec.rb
tom95: It's derived from the definitions in the factory. Are we expected to handle all of these values as constants? Since it's both test data, we duplicated values in both places all the time. Otherwise, we would need to carry around about ten constants to get meaningful tests.

app/models/date_range.rb
chrisma: How about a test that calls `human_attribute_name` with a parameter different from `start_date` or `end_date` ? That would reach the `super` statement in the control flow.
Maybe even better: try Rails' I18n : http://guides.rubyonrails.org/i18n.html#translations-for-active-record-models

Issue #21, opened 2016-11-10
US_1.6: Event drafts 🍪2


tobiduer:
**As** 
organizer 

**I want to** 
define draft events that are not published immediately 

**in order to** 
modify them together with my peers or colleagues. 

**5 🍪**

- [ ] I am able to click a save ("Speichern") button when I am done with my camp creation so that my camp is saved but not published
- [ ] When I am ready to publish my Draft i can click on a Publish ("Veröffentlichen") button and my Event becomes public. (on edit and creation page)

MasterCarl:
The model for workshop contains a field "active" (Boolean). Is this supposed to indicate whether a workshop is a draft?

chrisma:
This field is part of the starting application that was provided. Please feel free to delete/rename/modify any fields or classes as you see fit.

MasterCarl:
See pull request https://github.com/hpi-swt2/workshop-portal/pull/79

PR #79, opened 2016-11-28
21_1.6_EventDrafts

MasterCarl:
#21


bjrne:
The migrations are already outdated, please renew them before merging this.

MasterCarl:
I feel this should be a separate PR since it impacts other development operations

bjrne:
What do you mean with separate PR and "this"?

MasterCarl:
Never mind, I thought your comment was about https://github.com/hpi-swt2/workshop-portal/pull/80

spec/features/event_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    # The event should now be visible in the events list
+    visit events_path
+    expect(page).to have_text(@event.name)
+  end
+end
+
+def draft_button
+  find(:xpath, "//input[contains(@name, 'draft')]")
+end
```
tom95: As far as I understood capybara, you should always prefer to use `click_button "Veranstaltung veröffentlichen"`, because this resolves asynchronous operations, while find is instantaneous and may fail in more complicated scenarios (e.g. involving ajax and the like).

spec/features/event_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+
+def publish_button
+  find(:xpath, "//input[contains(@name, 'publish')]")
```
tom95: same as above

app/controllers/events_controller.rb
```diff
@@ -3,7 +3,7 @@ class EventsController < ApplicationController
 
   # GET /events
   def index
-    @events = Event.all
+    @events = Event.draft_is false
```
tom95: Is this a temporary solution? Right now, this is the only place I can access created events from to edit or, in particular, publish them. I guess the "proper" solution would be to toggle this based on user rights.

spec/features/event_spec.rb
```diff
@@ -0,0 +1,45 @@
+require "rails_helper"
+
+RSpec.feature "Draft events", :type => :feature do
```
tom95: During merging into `dev`, you may wanna move this either into its own file or as a `describe` within the enclosing `Event` block.

tom95 (changes requested):
The destiny of the `active` field is kinda unclear to me right now. I feel it should have been removed as part of this MR, but I'm not entirely sure either...

tom95 (approved):

PR #81, opened 2016-11-28
29 3.6 counter

frederike-ramin:
Add a counter of free and occupied places for an event.


Baschdl (dismissed):
Looks good. Please wait for approval of (full-time) developer ;)

db/migrate/20161103101916_create_applications.rb
```diff
@@ -3,7 +3,7 @@ def change
     create_table :applications do |t|
       t.string :motivation
       t.references :user, index: true, foreign_key: true
-      t.references :workshop, index: true, foreign_key: true
+      t.references :event, index: true, foreign_key: true
```
ekrebs5: (this is a question about our intern conventions, not about the code itself. This code should work with the new and the old version, I tested both on my system)
This migration was written when "event" was still called "workshop". Personally I think it should stay workshop in this method, otherwise we could rename all other migration methods and remove the "workshop->event" migration (This would however probably mean that everyone has to reset their database at least once). Do we have conventions/rules for this?

Baschdl: Yes, that's right. This renaming will probably crash the migration process. I'll talk with @bjrne from #62

bjrne: You should never touch old migration files (afaik). This is the old file that created the workshop column that later was renamed to event. I don't know what happens if you mess with this file and we need to execute it again. Please remove this change.

bjrne (dismissed):
Please remove the changes to the old migration file.

db/migrate/20161103101916_create_applications.rb
bjrne: @ekrebs5 I think the reason it both worked for you is that Ruby is so smart, it doesn't repeat migrations it already applied. Because the filename didn't change, this change was never tested when you just migrated your database. To really check this, you'd have to manually reverse all applied migrations until then, an migrate everything again. I am pretty sure the renaming would fail with this change, because there would be no more ":workshop"-column.

bjrne: Look at [this](http://edgeguides.rubyonrails.org/active_record_migrations.html#changing-existing-migrations) for more details: 
"In general, editing existing migrations is not a good idea. You will be creating extra work for yourself and your co-workers and cause major headaches if the existing version of the migration has already been run on production machines."

db/schema.rb
```diff
@@ -26,6 +26,15 @@
   add_index "application_letters", ["event_id"], name: "index_application_letters_on_event_id"
   add_index "application_letters", ["user_id"], name: "index_application_letters_on_user_id"
 
+  create_table "application_notes", force: :cascade do |t|
```
bjrne: Why do you have changes in this file when you didn't create any migrations? Have a loot at [this page](http://edgeguides.rubyonrails.org/active_record_migrations.html#schema-dumping-and-you) to get a better understanding of schema files if you didn't just forget to commit the migration file 😀

app/views/events/show.html.erb
```diff
@@ -23,4 +23,4 @@
 <% end %>
 
 <%= link_to t('.back', :default => t("helpers.links.back")),
-              events_path, :class => 'btn btn-default'  %>
```
bjrne: Is this whitespace change necessary?

app/controllers/application_letters_controller.rb
```diff
@@ -57,6 +57,6 @@ def set_application
 
     # Only allow a trusted parameter "white list" through.
     def application_params
-      params.require(:application_letter).permit(:motivation, :event_id, :status)
+      params.require(:application_letter).permit(:motivation, :user_id, :event_id, :status)
```
bjrne: I think this was removed a couple PRs earlier. I don't know your code that well, I just want to make sure this is not a merge error. Please give it a second look 🙂

aloeser: do you know why it was removed? An application letter without an user seems not really useful to me

frederike-ramin: My guess is, that you shoulndn't be able to create a letter with a manually set user. instead, in the create function, the user_id should be set by the backend itself.

cmfcmf: > My guess is, that you shoulndn't be able to create a letter with a manually set user. instead, in the create function, the user_id should be set by the backend itself.

Exactly. 😄

aloeser: @cmfcmf are there already tests for this? if yes, we also broke the tests.
If no, maybe your group should add some, since this seems like an important security feature :)

app/views/events/show.html.erb
aloeser: fixed

ekrebs5 (approved):
I think this is now ok :)

spec/features/events_spec.rb
```diff
@@ -1,21 +1,46 @@
...
+    expect(page).to have_text(I18n.t "free_places", count: @event.max_participants - 1, scope: [:events, :applicants_overview])
+    expect(page).to have_text(I18n.t "occupied_places", count: 1, scope: [:events, :applicants_overview])
+    # test negative amount of free places
+    @pupil = FactoryGirl.create(:profile)
+    @pupil.user.role = :pupil
+    @application_letter = FactoryGirl.create(:application_letter_accepted, event: @event, user: @pupil.user)
+    visit event_path(@event)
+    expect(page).to have_text(I18n.t "free_places", count: @event.max_participants - 2, scope: [:events, :applicants_overview])
+    expect(page).to have_text(I18n.t "occupied_places", count: 2, scope: [:events, :applicants_overview])
+  end
```
querenker: Maybe a loop could improve the readability of this test

querenker (approved):

PR #84, opened 2016-11-29
15 2.3 hide private data

hefowe:
Added error handling for CanCanCan errors to show the user nice error messages.


cmfcmf (approved):

config/locales/en.yml
```diff
@@ -21,6 +21,9 @@
 
 en:
   hello: "Welcome to the Workshop-Portal"
+  unauthorized:
+      manage:
+        all: "Not authorized to %{action} %{subject}."
```
cmfcmf: I think this should be changed to simply "You are not authorized to take this action." or "You are not authorized to view this page." without leaking the subject and action to the user.

cmfcmf (changes requested):
LGTM except for the alert message.

cmfcmf (approved):

PR #86, opened 2016-11-29
24 3.4 notes on applications additional tests

aloeser:
Added tests for routing and abilities.


spec/models/ability_spec.rb
```diff
@@ -71,6 +71,22 @@
     end
   end
 
+  %i[tutor organizer].each do |role|
+    it "can view and add notes to application letters" do
```
frederike-ramin: This doesn't seem to be recommended as it was not used in the other tests above. (also "{role}" is not that meaningful, is it?)

frederike-ramin: @querenker Would you approve that in that case?

querenker: @frederike-ramin Sorry for the late response (and the missing hash symbol). I meant
`it "can view and add notes to application letters as #{role}" do`. If one out of the two checked roles are failing the test, you will get this error message:

>  1) User can view and add notes to application letters
>       Failure/Error: expect(ability).to be_able_to(:view_and_add_notes, ApplicationLetter)
>         expected to be able to view_and_add_notes ApplicationLetter
>         ./spec/models/ability_spec.rb:79:in `block (3 levels) in <top (required)>'

Now I don't know which of the two roles is responsible for this error. If both role abilities are wrong you will get two times exactly the same error message (which is also weird for an other developer). With the change the error message would include the role which is responsible for the error.

>  1) User can view and add notes to application letters as tutor
>       Failure/Error: expect(ability).to be_able_to(:view_and_add_notes, ApplicationLetter)
>         expected to be able to view_and_add_notes ApplicationLetter
>         ./spec/models/ability_spec.rb:79:in `block (3 levels) in <top (required)>'

It is not a major issue and it concerns other (older) tests written like that, but I think it is nice and can maybe help other developers one day? :)

frederike-ramin: I see what you mean. I didn't expect the {role} to evaluate to the actual role, so I didn't get your point. I will change that.

querenker (approved):

ekrebs5:
I also approve this :)

ekrebs5 (approved):

PR #89, opened 2016-11-29

querenker:
Adds the functionality for organizer to change a application status with radio buttons. Tutors/Coaches can see the current statuses but they can't change these.


bjrne:
Please comment why you close pull requests so everybody knows the reason. Was it a duplicate, will you address it later?

cmfcmf:
> @querenker Please comment why you close pull requests, so everybody knows the reason. Was it a duplicate, will you address it later?

I assume because the target was set to `master` by accident. The follow-up PR can be found here: https://github.com/hpi-swt2/workshop-portal/pull/90

querenker:
@bjrne 
@cmfcmf is right. Meanwhile I also know that the base can be changed later on GitHub.

PR #90, opened 2016-11-29

querenker:
Adds functionality for organizer to change the application status with radio buttons.
Coaches/Tutors can only see the current application statuses.


app/controllers/application_letters_controller.rb
```diff
@@ -36,8 +36,8 @@ def create
 
   # PATCH/PUT /applications/1
   def update
-    if @application_letter.update(application_params)
-      redirect_to @application_letter, notice: 'Application was successfully updated.'
+    if @application_letter.update_attributes(application_params)
+      redirect_to :back, notice: 'Application was successfully updated.'
```
aloeser: what happens with :back if no HTTP referrer is set?

app/views/application_letters/show.html.erb
```diff
@@ -30,50 +30,50 @@
...
-  <dt><strong> Radio Buttons (#45 US_3.11: Mark as accepted/rejected) </strong></dt>
+  <dt><%= render partial: 'application_letters/application_selective', locals: {application_letter: @application_letter} %></dt>
 </dl>
 
 <p>
   <%= @application_letter.motivation %>
 </p>
 
 <% if can? :view_and_add_notes, ApplicationLetter  %>
-  <dl class="dl-horizontal">
```
aloeser: do we really need all those whitespace changes?

db/schema.rb
```diff
@@ -11,21 +11,29 @@
...
 
+  create_table "application_notes", force: :cascade do |t|
+    t.text     "note"
+    t.integer  "application_letter_id"
+    t.datetime "created_at",            null: false
+    t.datetime "updated_at",            null: false
+  end
+
+  add_index "application_notes", ["application_letter_id"], name: "index_application_notes_on_application_letter_id"
+
```
aloeser: I guess you didnt really move the application_notes table yourself? we had it in PR #81 twice in this file, and deleted the upper one. Now the upper one is there and the lower one is gone. But it shouldnt really matter which one we finally take.

spec/models/user_spec.rb
```diff
@@ -47,8 +47,8 @@
     other_event = FactoryGirl.create(:event)
     current_event = FactoryGirl.create(:event)
 
-    other_application_letter = FactoryGirl.create(:application_letter, user: user, event: other_event, status: true)
-    current_application_letter = FactoryGirl.create(:application_letter, user: user, event: current_event, status: true)
+    other_application_letter = FactoryGirl.create(:application_letter, user: user, event: other_event, status: :accepted)
+    current_application_letter = FactoryGirl.create(:application_letter, user: user, event: current_event, status: :accepted)
```
aloeser: you could use :application_letter_accepted ?

spec/models/user_spec.rb
```diff
@@ -58,8 +58,8 @@
     other_event = FactoryGirl.create(:event)
     current_event = FactoryGirl.create(:event)
 
-    other_application_letter = FactoryGirl.create(:application_letter, user: user, event: other_event, status: false)
-    current_application_letter = FactoryGirl.create(:application_letter, user: user, event: current_event, status: false)
+    other_application_letter = FactoryGirl.create(:application_letter, user: user, event: other_event, status: :rejected)
+    current_application_letter = FactoryGirl.create(:application_letter, user: user, event: current_event, status: :rejected)
```
aloeser: How about an factory model for an rejected application?

spec/controllers/application_letters_controller_spec.rb
```diff
@@ -128,12 +132,6 @@
         expect(assigns(:application_letter)).to eq(application)
       end
 
-      it "redirects to the application" do
```
aloeser: Why delete this test/redirect back to the root URL?

app/controllers/application_letters_controller.rb
querenker: It redirects back to the current site. If a organizer visits the event overview page, he/she can mark many applications as accepted/rejected without going back to the overview page every time. The HTTP referred is only neccessary for the test.

app/views/application_letters/show.html.erb
querenker: The whitespace changes were not intended.

db/schema.rb
querenker: I didn't merge the schema file directly. This file was generated through db:migrate, after meging the other stuff. If it doesn't matter I will keep it as it is?

spec/models/user_spec.rb
querenker: Since we have already such a factory it can of course be used.

db/schema.rb
bjrne: Keeping it as it is (if it is the newest version available) is a good idea. However, you need to merge dev into this branch again (conflict in this exact file) and should probably reverse & run the migrations again, just to be sure :)

querenker: @bjrne Thanks for your feedback. I will do so.

config/locales/de.application_letters.yml
```diff
@@ -12,3 +12,8 @@ de:
       application_note:
         one: "Anmerkung"
         other: "Anmerkungen"
+
+  application_status:
+    accepted: "Angenommen"
+    rejected: "Abgelehnt"
+    pending: "Ausstehend"
```
frederike-ramin: Perhaps, the verbs should be lower case? But that's a minor problem.

spec/controllers/application_letters_controller_spec.rb
frederike-ramin: Did you also write a test for that? That it redirects to the overview page? (Can't find any here)

spec/factories/application_letters.rb
```diff
@@ -8,20 +8,27 @@
...
     event
-    status nil
   end
 
   factory :application_letter_accepted, class: :application_letter do
     motivation "MyString"
     user
     event
-    status true
+    status :accepted
```
frederike-ramin: Just for my unterstanding: where is that defined? (That :accepted means 1)

spec/features/events_spec.rb
```diff
@@ -34,7 +34,28 @@
...
+
+  scenario "logged in as Organizer I can change application status with radio buttons" do
+    login(:organizer)
+
+    @pupil = FactoryGirl.create(:profile)
+    @application_letter = FactoryGirl.create(:application_letter, event: @event, user: @pupil.user)
+    visit event_path(@event)
+    ApplicationLetter.statuses.keys.each do |new_status|
+      choose(I18n.t "application_status.#{new_status}")
+      ApplicationLetter.where(id: @application_letter.id, status: new_status)
```
frederike-ramin: are you actually testing, that the status has changed? I don't see that here as you don't do anything with the return value of that request?

spec/factories/application_letters.rb
aloeser: In  app/models/application_letter.rb

spec/factories/application_letters.rb
```diff
@@ -8,20 +8,27 @@
...
+#  status      :integer          not null
 #
 
 FactoryGirl.define do
   factory :application_letter do
     motivation "MyString"
     user
     event
-    status nil
   end
```
aloeser: Maybe rather :pending for the status then nil? since pending is the default state after an application letter was created.

app/controllers/application_letters_controller.rb
aloeser: How about http://stackoverflow.com/questions/771656/correctly-doing-redirect-to-back-in-ruby-on-rails-when-referrer-is-not-availabl

spec/features/events_spec.rb
querenker: There is indeed something missing. Unfortunately making these test "right" results for some reason in an error...

querenker: Now it magically works :) Don't know why it works now resp. why it doesn't work before...

aloeser (approved):
all changes are worked in :)

frederike-ramin (approved):
Great.

Issue #24, opened 2016-11-10
US_3.4: Take notes on applications


julisa99:
**As** 
organizer or coach

**I want to**
be able to take notes on applications which other coaches can see

**in order to**
inform about my favorites, experiences with the person

**estimation: 5**

## Tasks
- [x] add notes-field / edit page => ask Tobi
- [x] add notes attribute to application => dependency: Tobi
- [x] show saved notes
- [x] save notes
- [x] ensure access rights (dependency #23)
- [x] make it pretty
- [x] Documentation
- [x] alles auf deutsch
- [x] feature test

## Acceptance criteria

- When a coach or an organizer visit the details page, he / she can add a textual note to application that should be displayed on the details page. Notes are visible to coaches and organizers. 

- These notes should never be visible to the applicants. 

- Notes are persistent.

tobiduer:
Additional details:
These notes should NOT be visible to the pupils!

frederike-ramin:
@tobiduer in the page overview of a user (in corresponding EPIC) there is a button for "Notiz hinzufügen". We talked about it as if there was supposed to be just one note, that can be edited. What is the case?

tobiduer:
@frederike-ramin: This was a misunderstanding. Please allow multiple notes in order to allow many coaches to add notes. 
Furthermore, notes can be edited and removed by the author and organizers only.

frederike-ramin:
@tobiduer This is a lot different from what we expected and estimated. It wasn't mentioned that we should also edit and delete notes. I suggest that we turn that into another User Story as it would make things a lot more complicated.

tobiduer:
@frederike-ramin: Ok. I will create a new user story for this. The team can decide on whether to make it in this sprint or the next.

PR #96, opened 2016-11-29
20 2.4 profile information coaches can view others applications

aloeser:
From our understanding coaches/tutors should be able to view application letters of other users, which is prohibited by the current implementation.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -42,7 +42,7 @@ class Ability
     if user.role? :tutor
       # Tutors can view Applications for Event
       can [:view_applicants], Event
-      can [:view_and_add_notes], ApplicationLetter
+      can [:view_and_add_notes, :show], ApplicationLetter
     end
     if user.role? :organizer
       can [:index, :show], Profile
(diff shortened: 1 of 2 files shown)


cmfcmf:
👎 I don't think this is what's wanted. At least I could never look into application letters of users for the past Schülerklub events.

aloeser:
@cmfcmf have you checked #24?
It says:
**As**
organizer or coach

**I want to**
be able to take notes on applications which other coaches can see

For me, this means a coach has to be able to see all applications for an event, not just his own.

tobiduer:
Alex is right. Coaches must be able to see applications, but not the personal information (addresse, ... ) 
They should vote for applications, therefore they should be able to read the letter of motivation.

cmfcmf:
> Alex is right. Coaches must be able to see applications, but not the personal information (addresse, ... )
> They should vote for applications, therefore they should be able to read the letter of motivation.

Ah, I see! Thank you for clarifying, Tobi!

ekrebs5 (approved):

querenker (approved):

Issue #48, opened 2016-11-18
US_1.8: Choose Eventkind 🍪1


Helene0:
**As** 
organizer 

**I want to** 
be able to choose when I create a new event what kind of Event it will be (Camp/ Workshop)

**in order to** 
organize my events, and can start different processes

**5 🍪**

- [ ] I have a drop down menu from which I can choose the kind of event that mine is supposed to be

PR #105, opened 2016-11-30
48 1.8 choose kind of event

bjrne:
#48


Baschdl:
I think the update of db/seeds.rb is missing

app/views/events/show.html.erb
```diff
@@ -1,6 +1,6 @@
 <%- model_class = Event -%>
 <div class="page-header">
-  <h1><%=t '.title', :default => model_class.model_name.human.titleize %></h1>
+  <h1><%= @event.kind.humanize %> <%=t '.title', :default => model_class.model_name.human.titleize %></h1>
```
cmfcmf: This only works because "Workshop" and "Camp" are spelled the same in English and German.

spec/features/event_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+      expect(page).to have_text('Camp')
+    end
+
+    it "should allow picking the event kind workshop" do
+      visit new_event_path
+      fill_in 'Max participants', :with => 25
+      choose('Workshop')
+      click_button "Event erstellen"
+      expect(page).to have_text('Workshop')
+    end
```
cmfcmf: The above could be simplified by a loop like done for roles here: https://github.com/hpi-swt2/workshop-portal/blob/dev/spec/models/ability_spec.rb#L7

spec/features/event_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    it "should display the event kind" do
+      event = FactoryGirl.create(:event, kind: :workshop)
+      visit event_path(event)
+      expect(page).to have_text('Workshop')
+
+      event = FactoryGirl.create(:event, kind: :camp)
+      visit event_path(event)
+      expect(page).to have_text('Camp')
+    end
+  end
```
cmfcmf: The test above could also benefit from the same loop.

cmfcmf (commented):
I had some minor thoughts on this :)

cmfcmf:
> I think the update of db/seeds.rb is missing

Not necessarily, because `kind` has a default value.

tom95:
Thanks @cmfcmf, I added the loops. `kind.humanize` should be okay for this sprint, as translations were not a requirement so far

app/views/events/_form.html.erb
```diff
@@ -16,6 +16,18 @@
   <% end %>
 
   <div class="form-group">
+    <div class="btn-group col-lg-10 col-lg-offset-2" data-toggle="buttons">
+      <% Event.kinds.keys.each do |key| %>
+        <label class="btn btn-default <%= 'active' if @event.kind == key %>">
+          <%= f.radio_button :kind, key %>
+          <%= key.humanize %>
```
MasterCarl: Shouldn't the radio button be selected if the class 'active' is set for the label? Or am I missing something here?

MasterCarl: I just checked in my browser, it seems to work

app/views/events/show.html.erb
```diff
@@ -1,6 +1,6 @@
 <%- model_class = Event -%>
 <div class="page-header">
-  <h1><%= t '.title', :default => model_class.model_name.human.titleize %></h1>
+  <h1><%= @event.kind.humanize %> <%=t '.title', :default => model_class.model_name.human.titleize %></h1>
```
MasterCarl: I don't think that the event kind should be the first part of the title, but it's good to have it displayed

MasterCarl (changes requested):
This PR looks good so far, but I think the kind of event should also be displayed on the index page

bjrne:
@MasterCarl Isn't that what @LeaGe is working on?

LeaGe:
@bjrne yupp. :)

app/views/events/show.html.erb
tom95: I'll put it last, in parenthesis. So we would get "Mint Camp (Camp)" which is still kinda stupid, but at least explicit for the user.

MasterCarl (approved):

PR #107, opened 2016-11-30
22 3.2 send emails in a bundle

Lando-L:


app/models/event.rb
```diff
@@ -15,6 +15,32 @@ class Event < ActiveRecord::Base
 
   validates :max_participants, numericality: { only_integer: true, greater_than: 0 }
 
+  # Returns whether all application_letters are classified or not
+  #
+  # @param none
+  # @return [Boolean] if status of all application_letters is not equal 2
+  def applicationsClassified?
```
querenker: I think methods should be written in this form `applications_classified?`.

spec/features/events_spec.rb
```diff
@@ -22,6 +22,49 @@
...
+
+  scenario "logged in as Organizer I want to open a modal by clicking on sending emails" do
+    login(:organizer)
+    @event.update!(max_participants: 2)
+    @pupil1 = FactoryGirl.create(:profile)
+    @pupil1.user.role = :pupil
+    @pupil2 = FactoryGirl.create(:profile)
+    @pupil2.user.role = :pupil
+    @acceptedApplication = FactoryGirl.create(:application_letter, :event => @event, :user => @pupil1.user, :status => 1)
+    @acceptedApplication2 = FactoryGirl.create(:application_letter, :event => @event, :user => @pupil2.user, :status => 1)
```
aloeser: same as above with :application_letter_accepted

spec/models/event_spec.rb
```diff
@@ -20,6 +20,31 @@
     expect(event).to be_valid
   end
 
+  it "checks if there are unclassified applications_letters" do
+    event = FactoryGirl.create(:event)
+    acceptedApplicationLetter = FactoryGirl.create(:application_letter, :event => event, :user => FactoryGirl.create(:user), :status => 1)
```
aloeser: same with :application_letter_accepted

spec/models/event_spec.rb
```diff
@@ -20,6 +20,31 @@
...
+    event.application_letters.push(rejectedApplicationLetter)
+    expect(event.applicationsClassified?).to eq(false)
+  end
+
+  it "computes the email addreses of the accepted and the rejected applications" do
+    event = FactoryGirl.create(:event)
+    acceptedApplicationLetter1 = FactoryGirl.create(:application_letter, :event => event, :user => FactoryGirl.create(:user), :status => 1)
+    acceptedApplicationLetter2 = FactoryGirl.create(:application_letter, :event => event, :user => FactoryGirl.create(:user), :status => 1)
+    acceptedApplicationLetter3 = FactoryGirl.create(:application_letter, :event => event, :user => FactoryGirl.create(:user), :status => 1)
+    rejectedApplicationLetter = FactoryGirl.create(:application_letter, :event => event, :user => FactoryGirl.create(:user), :status => 0)
```
aloeser: :application_letter_rejected should exist too :)

spec/models/event_spec.rb
```diff
@@ -20,6 +20,31 @@
...
+    acceptedApplicationLetter = FactoryGirl.create(:application_letter, :event => event, :user => FactoryGirl.create(:user), :status => 1)
+    event.application_letters.push(acceptedApplicationLetter)
+    expect(event.applicationsClassified?).to eq(true)
+
+    rejectedApplicationLetter = FactoryGirl.create(:application_letter, :event => event, :user => FactoryGirl.create(:user), :status => 2)
+    event.application_letters.push(rejectedApplicationLetter)
+    expect(event.applicationsClassified?).to eq(false)
+  end
+
+  it "computes the email addreses of the accepted and the rejected applications" do
```
aloeser: typo

spec/models/event_spec.rb
```diff
@@ -20,6 +20,31 @@
     expect(event).to be_valid
   end
 
+  it "checks if there are unclassified applications_letters" do
+    event = FactoryGirl.create(:event)
+    acceptedApplicationLetter = FactoryGirl.create(:application_letter, :event => event, :user => FactoryGirl.create(:user), :status => 1)
+    event.application_letters.push(acceptedApplicationLetter)
+    expect(event.applicationsClassified?).to eq(true)
+
+    rejectedApplicationLetter = FactoryGirl.create(:application_letter, :event => event, :user => FactoryGirl.create(:user), :status => 2)
```
aloeser: I think the name rejectedApplicationLetter is misleading, isnt it rather an application letter which is still pending (since you use status: 2)?

app/models/event.rb
```diff
@@ -15,6 +15,32 @@ class Event < ActiveRecord::Base
...
+    application_letters.all? { |application_letter| application_letter.status != 'pending' }
+  end
+
+  # Returns a string of all email addresses of rejected applications
+  #
+  # @param none
+  # @return [String] Concatenation of all email addresses of rejected applications, seperated by ','
+  def compute_rejected_applications_emails
```
querenker: I think the name doesn't fit very well. "Compute" sounds for me like a time consuming task with a fancy algorithm. Moreover this method doesn't compute/generate the rejected emails. But it is indeed not so easy to find a proper name...

app/models/event.rb
```diff
@@ -15,6 +15,32 @@ class Event < ActiveRecord::Base
...
+    rejected_applications = application_letters.select { |application_letter| application_letter.status == 'rejected' }
+    rejected_applications.map{ |applications_letter| applications_letter.user.email }.join(',')
+  end
+
+  # Returns a string of all email addresses of accepted applications
+  #
+  # @param none
+  # @return [String] Concatenation of all email addresses of accepted applications, seperated by ','
+  def compute_accepted_applications_emails
```
querenker: Analogous to `compute_rejected_applications_emails`

spec/views/events/show.html.erb_spec.rb
```diff
@@ -26,6 +26,14 @@
     expect(rendered).to have_text(I18n.t 'occupied_places', count: occupied_places, scope: [:events, :applicants_overview])
   end
 
+  it "displays counter" do
+    free_places = assign(:free_places, @event.compute_free_places)
+    occupied_places = assign(:occupied_places, @event.compute_occupied_places)
+    render
```
aloeser: This looks like a merging mistake to me, i think that test is duplicated now (if you view the whole file it is there twice)

aloeser (changes requested):

app/views/events/_applicants_overview.html.erb
```diff
@@ -41,18 +40,54 @@
...
+          events_path, :class => 'btn btn-default'%>
     </div>
     <div class="btn-group pull-right" role="group">
-      <%= link_to "Zusagen verschicken",
-                  events_path, :class => 'btn btn-default'%>
-
-      <%= link_to "Absagen verschicken",
-                  events_path, :class => 'btn btn-default'%>
+      <% if (!@event.applicationsClassified?) %>
+          <div class="tooltip-wrapper disabled has-tooltip" data-toggle="tooltip" title="Bewerbung(en) wurden noch nicht klassifiziert">
```
querenker: Please use locals for language dependent strings.

db/seeds.rb
```diff
@@ -103,7 +103,7 @@
     vegan: false,
     allergic: false,
     allergies: "",
-    status: nil,
+    status: 2,
```
querenker: Can you use `status: :pending` or  `ApplicationLetter.statuses[:pending]`? I think this would improve readability.

spec/features/events_spec.rb
```diff
@@ -22,6 +22,49 @@
     expect(page).to have_css("div#occupied_places")
   end
 
+  scenario "logged in as Organizer I want to be unable to send emails if there is any unclassified application left" do
+    login(:organizer)
+    @event.update!(max_participants: 1)
+    @pupil1 = FactoryGirl.create(:profile)
```
querenker: I think you can remove the 1 if only one pupil gets created :)

spec/features/events_spec.rb
```diff
@@ -22,6 +22,49 @@
...
+    @pupil1.user.role = :pupil
+    @nilApplication = FactoryGirl.create(:application_letter, :event => @event, :user => @pupil1.user, :status => 2)
```
querenker: Why is it called `nilApplication`? I think something like `pending_application` would be clearer?

querenker: General question: did we have any standard for `:event => @event` vs `event: @event`. Are there any distinctions?

querenker: Please use the more readable` :status => :pending` instead of `:status => 2`

spec/features/events_spec.rb
```diff
@@ -22,6 +22,49 @@
...
+
+  scenario "logged in as Organizer I want to be unable to send emails if there is a negative number of free places left" do
+    login(:organizer)
+    @event.update!(max_participants: 1)
+    @pupil1 = FactoryGirl.create(:profile)
+    @pupil1.user.role = :pupil
+    @pupil2 = FactoryGirl.create(:profile)
+    @pupil2.user.role = :pupil
+    @acceptedApplication = FactoryGirl.create(:application_letter, :event => @event, :user => @pupil1.user, :status => 1)
+    @acceptedApplication2 = FactoryGirl.create(:application_letter, :event => @event, :user => @pupil2.user, :status => 1)
```
querenker: `:status => :accepted` rsp. `status: :accepted` (same for many other positions). In this case you can even use the [application_letter_accepted factory](https://github.com/hpi-swt2/workshop-portal/blob/dev/spec/factories/application_letters.rb#L29)

spec/features/events_spec.rb
```diff
@@ -22,6 +22,49 @@
...
+    visit event_path(@event)
+    click_button "Zusagen verschicken"
+    expect(page).to have_selector('div', :id => 'send-emails-modal')
+    #expect(find('#send-emails-list', :visible => false).value).to eq(@event.compute_accepted_applications_emails)
```
querenker: dead code

app/assets/javascripts/events.js
```diff
@@ -1,2 +1,30 @@
 // Place all the behaviors and hooks related to the matching controller here.
 // All this logic will automatically be available in application.js.
+
+jQuery(function() {
+
+    $('#send-emails-modal').on('show.bs.modal', function (event) {
```
querenker: What is modal?

querenker (changes requested):
Otherwise it looks good for me. I can't say very much about the JavaScript part and have only scanned the tests. At least the tests I will check later, but at first glance it looks good for me.

spec/features/events_spec.rb
bjrne: As far as i know, the two are exactly the same, the second one is the "more recent" syntax. But I don't think we formally agreed to standards.

db/migrate/20161103101916_create_applications.rb
```diff
@@ -3,7 +3,7 @@ def change
     create_table :applications do |t|
       t.string :motivation
       t.references :user, index: true, foreign_key: true
-      t.references :workshop, index: true, foreign_key: true
+      t.references :event, index: true, foreign_key: true
```
cmfcmf: > Do we really need an additional migration for this?

I don't think so. I assume @Baschdl's comment was more of a general comment for renaming fields. This field however **is** already renamed, so changing the migration is not necessary and should be reverted.

bjrne: Please PLEASE read [this](http://edgeguides.rubyonrails.org/active_record_migrations.html#changing-existing-migrations), especially chapters 4, 5 & 6! @Lando-L

app/assets/javascripts/events.js
Lando-L: It is bootstrap component http://getbootstrap.com/javascript/#modals. But this will be replaced from another view next sprint.

app/models/event.rb
```diff
@@ -15,6 +15,32 @@ class Event < ActiveRecord::Base
...
+  def applicationsClassified?
+    application_letters.all? { |application_letter| application_letter.status != 'pending' }
+  end
+
+  # Returns a string of all email addresses of rejected applications
+  #
+  # @param none
+  # @return [String] Concatenation of all email addresses of rejected applications, seperated by ','
+  def compute_rejected_applications_emails
+    rejected_applications = application_letters.select { |application_letter| application_letter.status == 'rejected' }
```
Lando-L: I actually don't know the difference, but the .where() version is shorter ;)

cmfcmf: I think the difference is that `select` fetches all objects from the database and then iterates over them to select the right ones, whereas `where` uses SQL to and thus does the selection inside the db.

app/models/event.rb
```diff
@@ -18,26 +18,27 @@ class Event < ActiveRecord::Base
...
 
   # Returns a string of all email addresses of rejected applications
   #
   # @param none
   # @return [String] Concatenation of all email addresses of rejected applications, seperated by ','
-  def compute_rejected_applications_emails
-    rejected_applications = application_letters.select { |application_letter| application_letter.status == 'rejected' }
+  def rejected_applications_emails
+    #rejected_applications = application_letters.select { |application_letter| application_letter.status == 'rejected' }
+    rejected_applications = application_letters.where(status: ApplicationLetter.statuses[:rejected])
```
aloeser: just for my understanding, is application_letters.x() the same model access like ApplicationLetter.x()?

app/views/events/_applicants_overview.html.erb
```diff
@@ -61,8 +61,8 @@
             <%= button_tag "Absagen verschicken", type: 'button', class: 'send-emails-button btn btn-default', disabled: true %>
           </div>
       <% else %>
-          <%= button_tag "Zusagen verschicken", type: 'button', class: 'btn btn-default', data: {toggle: 'modal', target: '#send-emails-modal', title: 'Zusagen verschicken', list: @event.compute_accepted_applications_emails} %>
-          <%= button_tag "Absagen verschicken", type: 'button', class: 'btn btn-default', data: {toggle: 'modal', target: '#send-emails-modal', title: 'Absagen verschicken', list: @event.compute_rejected_applications_emails} %>
+          <%= button_tag "Zusagen verschicken", type: 'button', class: 'btn btn-default', data: {toggle: 'modal', target: '#send-emails-modal', title: 'Zusagen verschicken', list: @event.accepted_applications_emails} %>
+          <%= button_tag "Absagen verschicken", type: 'button', class: 'btn btn-default', data: {toggle: 'modal', target: '#send-emails-modal', title: 'Absagen verschicken', list: @event.rejected_applications_emails} %>
```
aloeser: We should use locales (i18n) for "Zusagen/Absagen verschicken"

spec/features/events_spec.rb
```diff
@@ -22,6 +22,45 @@
...
 
+  scenario "logged in as Organizer I want to be unable to send emails if there is any unclassified application left" do
+    login(:organizer)
+    @event.update!(max_participants: 1)
+    @pupil = FactoryGirl.create(:profile)
+    @pupil.user.role = :pupil
+    @pending_application = FactoryGirl.create(:application_letter, :event => @event, :user => @pupil.user)
+    visit event_path(@event)
+    expect(page).to have_button("Zusagen verschicken", disabled: true)
+    expect(page).to have_button("Absagen verschicken", disabled: true)
```
aloeser: I suggest using i18n in the tests too

spec/features/events_spec.rb
```diff
@@ -22,6 +22,45 @@
...
+    @event.update!(max_participants: 1)
+    @pupil1 = FactoryGirl.create(:profile)
+    @pupil1.user.role = :pupil
+    @pupil2 = FactoryGirl.create(:profile)
+    @pupil2.user.role = :pupil
+    @accepted_application = FactoryGirl.create(:application_letter_accepted, :event => @event, :user => @pupil1.user)
+    @accepted_application_2 = FactoryGirl.create(:application_letter_accepted, :event => @event, :user => @pupil2.user)
+    visit event_path(@event)
+    expect(page).to have_button("Zusagen verschicken", disabled: true)
+    expect(page).to have_button("Absagen verschicken", disabled: true)
```
aloeser: i18n again

spec/features/events_spec.rb
```diff
@@ -22,6 +22,45 @@
...
+    login(:organizer)
+    @event.update!(max_participants: 2)
+    @pupil1 = FactoryGirl.create(:profile)
+    @pupil1.user.role = :pupil
+    @pupil2 = FactoryGirl.create(:profile)
+    @pupil2.user.role = :pupil
+    @accepted_application = FactoryGirl.create(:application_letter_accepted, :event => @event, :user => @pupil1.user)
+    @accepted_application_2 = FactoryGirl.create(:application_letter_accepted, :event => @event, :user => @pupil2.user)
+    visit event_path(@event)
+    click_button "Zusagen verschicken"
```
aloeser: i18n again

app/views/events/_applicants_overview.html.erb
```diff
@@ -41,18 +40,54 @@
...
     </div>
     <div class="btn-group pull-right" role="group">
-      <%= link_to "Zusagen verschicken",
-                  events_path, :class => 'btn btn-default'%>
-
-      <%= link_to "Absagen verschicken",
-                  events_path, :class => 'btn btn-default'%>
+      <% if (!@event.applications_classified?) %>
+          <div class="tooltip-wrapper disabled has-tooltip" data-toggle="tooltip" title="Bewerbung(en) wurden noch nicht klassifiziert">
+            <%= button_tag "Zusagen verschicken", type: 'button', class: 'send-emails-button btn btn-default', disabled: true %>
```
aloeser: Please use i18n

aloeser: pretty much for the whole file, there shouldnt be a single german text in there

aloeser (changes requested):
Changes look good to me, just needs i18n

app/models/event.rb
aloeser: Sorry, my bad, i thought it were code for a view. For a model i totally agree :)

app/assets/javascripts/events.js
querenker: Thanks for the explanation :)

aloeser (commented):
changes look good to me, still need to test it locally

aloeser (approved):
locally tested, works! You will need to merge dev into this though

app/views/events/_applicants_overview.html.erb
```diff
@@ -41,18 +40,54 @@
...
+
+<!-- Bootstrap Modal for sending emails -->
+<div id="send-emails-modal" class="modal fade" role="dialog">
+  <div class="modal-dialog">
+
+    <!-- Modal content-->
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal">&times;</button>
+        <h4 class="modal-title">Emails verschicken</h4>
```
querenker: i18n locale missing

app/views/events/_applicants_overview.html.erb
```diff
@@ -41,18 +40,54 @@
...
+      </div>
+      <div class="modal-body">
+        <%= link_to "Senden", '#', :id => 'send-emails-mailto', :class => 'btn btn-default' %>
+        <%= button_tag "In die Zwischenablage kopieren", type: 'button', id: 'send-emails-clipboard', class: 'btn btn-default' %>
```
querenker: i18n for "Senden" as well as "In die Zwischenablage kopieren"

spec/features/events_spec.rb
```diff
@@ -22,6 +22,45 @@
...
+
+  scenario "logged in as Organizer I want to be unable to send emails if there is a negative number of free places left" do
+    login(:organizer)
+    @event.update!(max_participants: 1)
+    @pupil1 = FactoryGirl.create(:profile)
+    @pupil1.user.role = :pupil
+    @pupil2 = FactoryGirl.create(:profile)
+    @pupil2.user.role = :pupil
+    @accepted_application = FactoryGirl.create(:application_letter_accepted, :event => @event, :user => @pupil1.user)
+    @accepted_application_2 = FactoryGirl.create(:application_letter_accepted, :event => @event, :user => @pupil2.user)
```
querenker: Maybe this can be put in a loop? Something like:
```
2.times do
  pupil = FactoryGirl.create(:profile)
  pupil.user.role = :pupil
  FactoryGirl.create(:application_letter_accepted, :event => @event, :user => @pupil.user)
end
```

spec/features/events_spec.rb
```diff
@@ -22,6 +22,45 @@
...
+
+  scenario "logged in as Organizer I want to open a modal by clicking on sending emails" do
+    login(:organizer)
+    @event.update!(max_participants: 2)
+    @pupil1 = FactoryGirl.create(:profile)
+    @pupil1.user.role = :pupil
+    @pupil2 = FactoryGirl.create(:profile)
+    @pupil2.user.role = :pupil
+    @accepted_application = FactoryGirl.create(:application_letter_accepted, :event => @event, :user => @pupil1.user)
+    @accepted_application_2 = FactoryGirl.create(:application_letter_accepted, :event => @event, :user => @pupil2.user)
```
querenker: See [previous test](https://github.com/hpi-swt2/workshop-portal/pull/107/files#diff-f20a1b372768b0b5eebe83bc6aa154f8R36) If this snippet is more frequently used it could be extracted in a separate method, but I think that is currently not essential.

spec/models/event_spec.rb
```diff
@@ -21,6 +21,31 @@
...
+  it "computes the email addresses of the accepted and the rejected applications" do
+    event = FactoryGirl.create(:event)
+    accepted_application_letter_1 = FactoryGirl.create(:application_letter_accepted, :event => event, :user => FactoryGirl.create(:user))
+    accepted_application_letter_2 = FactoryGirl.create(:application_letter_accepted, :event => event, :user => FactoryGirl.create(:user))
+    accepted_application_letter_3 = FactoryGirl.create(:application_letter_accepted, :event => event, :user => FactoryGirl.create(:user))
+    rejected_application_letter = FactoryGirl.create(:application_letter_rejected, :event => event, :user => FactoryGirl.create(:user))
+    event.application_letters.push(accepted_application_letter_1)
+    event.application_letters.push(accepted_application_letter_2)
+    event.application_letters.push(accepted_application_letter_3)
+    event.application_letters.push(rejected_application_letter)
```
querenker: Maybe here is also a loop possible?

querenker (changes requested):
As far as I can judge there are no critical issues left. Just some minor suggestions.

app/views/events/_applicants_overview.html.erb
Lando-L: This is never shown. This gets either overridden by '.sending_acceptances' or '.sending_rejections'. Should I add a locale for it anyway?

app/views/events/_applicants_overview.html.erb
Lando-L: We are writing an email template in the current sprint and discard this modal. So i will leave it like this, because we will certainly need an own `locals.de.emails.yml` and remember this one!

db/migrate/20161103101916_create_applications.rb
Baschdl: No, we don't need a extra migration for this. It was a general comment.

app/models/event.rb
```diff
@@ -78,6 +78,33 @@ def agreement_letter_for(user)
...
+  def applications_classified?
+    application_letters.all? { |application_letter| application_letter.status != 'pending' }
+  end
+
+  # Returns a string of all email addresses of rejected applications
+  #
+  # @param none
+  # @return [String] Concatenation of all email addresses of rejected applications, seperated by ','
+  def email_adresses_of_rejected_applicants
+    #rejected_applications = application_letters.select { |application_letter| application_letter.status == 'rejected' }
```
querenker: Dead code

app/views/events/_applicants_overview.html.erb
querenker: Can the Tag be empty or filled in the English Version? I think the Code should be consitently in English.

spec/features/events_spec.rb
querenker: Just for my interest Is there a reason for not using a loop?

querenker (changes requested):

spec/models/event_spec.rb
```diff
@@ -59,6 +59,28 @@
...
+    expect(event.applications_classified?).to eq(false)
+  end
+
+  it "computes the email addresses of the accepted and the rejected applications" do
+    event = FactoryGirl.create(:event)
+    accepted_application_letter_1 = FactoryGirl.create(:application_letter_accepted, :event => event, :user => FactoryGirl.create(:user))
+    accepted_application_letter_2 = FactoryGirl.create(:application_letter_accepted, :event => event, :user => FactoryGirl.create(:user))
+    accepted_application_letter_3 = FactoryGirl.create(:application_letter_accepted, :event => event, :user => FactoryGirl.create(:user))
+    rejected_application_letter = FactoryGirl.create(:application_letter_rejected, :event => event, :user => FactoryGirl.create(:user))
+    [accepted_application_letter_1, accepted_application_letter_2, accepted_application_letter_3, rejected_application_letter].each { |letter| event.application_letters.push(letter) }
```
querenker: Can this line be replaced by
`event.application_letters += [accepted_application_letter_1, accepted_application_letter_2, accepted_application_letter_3, rejected_application_letter]`? As further improvement can the applications can be stored in an array with an descriptive name?

querenker (approved):

Issue #23, opened 2016-11-10
US_3.3: Applicants overview


julisa99:
**As** 
organizer

**I want to**
have an overview with all applicants (personal data, how often already applied, how often accepted / rejected)

**in order to**
give new rejected applicants a new chance. 

**estimation: 13**
**new estimation: 1**

## Tasks
- [x] Table for Workshop Applicants
- [x] Buttons for other required features
- [x] Add gender and age
- [x] Add number of participation to User 
- [x] Add application_letter details page
- [x] Polish interface design
- [x] Manage visibility options (e.g. not for pupil)

## Acceptance criteria
- **When a user logs into the workshop portal, and selects one workshop, he / she should see this overview page.**
![overview](https://cloud.githubusercontent.com/assets/16103359/20379737/31f754ea-ac9e-11e6-8d28-3c23ac49d4a4.jpg)
See [UI wireframe for overview](https://wireframe.cc/Roqo7E)

- **When the user clicks on "details", the following page is displayed:**
![detail](https://cloud.githubusercontent.com/assets/16103359/20379736/31d8bc56-ac9e-11e6-9e42-d362850ead99.jpg)
See [UI wireframe for details](https://wireframe.cc/86Ceeg)

- **Pupils can not see these two pages.**

- **Coaches and organizers see all applications of the opened workshop.**

- **Data of the current data base state is visualized correctly.**

tobiduer:
See https://wireframe.cc/Roqo7E

tobiduer:
# Overview
![overview](https://cloud.githubusercontent.com/assets/16103359/20379737/31f754ea-ac9e-11e6-8d28-3c23ac49d4a4.jpg)

# Details 
![detail](https://cloud.githubusercontent.com/assets/16103359/20379736/31d8bc56-ac9e-11e6-9e42-d362850ead99.jpg)

tobiduer:
# Details 
![detail](https://cloud.githubusercontent.com/assets/16103359/20379736/31d8bc56-ac9e-11e6-9e42-d362850ead99.jpg)

tobiduer:
Feedback on the current state on (herko)[http://workshop-portal-dev.herokuapp.com]:

- [ ] Please remove the issue numbers and texts in order to provide a good feeling of the view.

- [ ] Please move the back button to the left.

- [ ] Please swap the accept all and print all button positions in order to provide a chronological flow from left to right.

[...]

PR #109, opened 2016-12-01
23 3.3 applicants overview, typo/bug fix

ekrebs5:
application letter detail page should show "Bewerbung für", since it is the page for the application letter, not of the applicant

the plural of "Bewerbung" is "Bewerbungen", not "Bewerbungs"

see comments in issue #23


querenker (approved):

app/views/application_letters/index.html.erb
```diff
@@ -1,6 +1,6 @@
 <%- model_class = ApplicationLetter -%>
 <div class="page-header">
-  <h1><%=t '.title', :default => model_class.model_name.human.pluralize.titleize %></h1>
+  <h1><%=t '.title', :default => model_class.model_name.human(:count => model_class.count).titleize %></h1>
```
aloeser: I think right now it takes the number of all application letters in order to decide between singular and plural. Is this intentioned?

ekrebs5: It is what I intended at least.

I talked with our PO @tobiduer today and he said that he does not care about this detail, so this shpuld be fine.
Apparently, the POs are talking about this page anyway, since students should only see their own application letters and all others only need to see them as part of an event. But this is not part of any user story right now, it is currently discussed. Ergo this is not part of our acceptamce criteria.

frederike-ramin (approved):

app/views/application_letters/index.html.erb
aloeser: Well I think it will lead to an bug report, as soon as this page gets used. But I think you can merge it anyway

ekrebs5: Well, it should be a new issue.

Our issue originally only concerned the event overview page and the application letter detail page. Then we also got a bug issue to turn Berwerbungs into the correct Bewerbungen which does concern the application letter index page (as in if have the correct prural in our data anyway we can also use it in this page), but that issue was also not about restructuring the application letter index page.

PR #113, opened 2016-12-02
Test if seeding the database works

cmfcmf:
Anyone please review and merge this :)
(@hpi-swt2/team-d-16, @hpi-swt2/team-i-16, @hpi-swt2/team-k-16, @hpi-swt2/team-r-16)

PR #114, opened 2016-12-02
Fix seeds not containing dateranges and setting application status to nil

tom95:
This adds date ranges to an event so that validations pass. Also, application_letter used to set status to nil, failing validation as well. This branch fixes both errors.


cmfcmf:
I already added this as part of #113. But we can also merge your PR and then merge mine with your changes included, if you want.

cmfcmf (approved):

PR #132, opened 2016-12-03
Add "draft" to db/schema.rb

bjrne:
Somehow the most recent version on dev doesn't have it.


tom95:
I think this should only go in once https://github.com/hpi-swt2/workshop-portal/pull/79 is merged (which also contains the schema.rb change) or am I mixing something up?

bjrne:
Oh, i thought it was already merged, by bad. But somehow i got this change only running migrations that are already on dev...

Issue #17, opened 2016-11-10
US_1.3: Time span for events 🍪1


tobiduer:
**As** 
organizer 

**I want to** 
enter a time span for events 

**in order to** 
inform pupils and others about the date (and see the dates in the calender)

**40 🍪** 

Acceptance Criteria
 - [ ] I can't choose dates in the past
 - [ ] I can't choose unreasonable timespans in terms of length (warning, but allowed)
 - [ ] I can enter multiple dates
 - [ ] I can't have an enddate before a startdate
 - [ ] I get error messages if any of the above cases happen
 - [ ] It is clear to me which format I have to use to input my dates 
 - [ ] if its only one da i only want to see one date

PR #134, opened 2016-12-03
17 1.3 js fix

tom95:
This fixes adding timespans in #17 and finally enables javascript tests.

**NOTE:** In order for javascript tests to work, you'll need to install phantomjs

**Ubuntu**: `apt-get install phantomjs`  
**Mac**: `brew install phantomjs`  
**Windows**: https://bitbucket.org/ariya/phantomjs/downloads/phantomjs-1.9.8-windows.zip 

If you encounter problems, detailed instructions can be found here: https://github.com/teampoltergeist/poltergeist


Vagrantfile
```diff
@@ -22,4 +22,9 @@ Vagrant.configure("2") do |config|
   # port forward
   config.vm.network :forwarded_port, host: 3000, guest: 3000
   config.vm.synced_folder ".", "/home/vagrant/hpi-swt2"
+
+  config.vm.provision "shell", inline: <<-SHELL
+    apt-get update
+    apt-get install -y phantomjs
+  SHELL
 end
```
cmfcmf: @tom95 I have also added code to install PhantomJS with Vagrant. People using Vagrant can simply do `vagrant reload --provision`.

spec/rails_helper.rb
```diff
@@ -58,7 +58,7 @@
   # If you're not using ActiveRecord, or you'd prefer not to run each of your
   # examples within a transaction, remove the following line or assign false
   # instead of true.
-  config.use_transactional_fixtures = true
+  # config.use_transactional_fixtures = true
```
MasterCarl: I'd rather set this to `false`

MasterCarl (approved):
I tested it locally and it works. Good to have JavaScript tests!

spec/rails_helper.rb
tom95: It is being set to false in a different place already (spec/spec_helper.rb). It made more sense for me to do that there since the comment explaining why it's false is over there as well. Otherwise I would agree :)

PR #135, opened 2016-12-03
Add rake task db:populate_sample_data

### LINKED ISSUE #172: Move sample data from seeds.rb to somewhere else

bjrne:
You can now easily add more data to your database. I already added some events, feel free to add more samples!

Apparently, seeds.rb is only for a bare minimum of data. It should contain just enough data so the platform runs. This is the best way I found for conveniently adding more data the rails way. Feedback welcome!

/cc @Baschdl @h0rnyc0rny @nilsthamm 

closes #172


bjrne:
Thanks for your feedback @MasterCarl, the SMs discussed this a little further and various changes will be addressed before this is merged, but we will definitely add this soon! :)

bjrne:
I guess this only makes sense with more data. At the moment all the data is only a duplicate of the seeds which would make things a little confusing...

bjrne:
@nilsthamm I don't think we should remove the data from seeds.rb. I'd only add more data, since the data in seeds is necessary for the platform to work (or isn't it)? @hpi-swt2/scrum-master-2016

nilsthamm:
As I see it, it works fine without all the data in seeds.rb. Things like events application letter and users unlike admins are only necessary to present functionalities to the customer. So this would be sample data in my opinion.

bjrne:
Hm, that's actually right. I wonder why we had this data in seeds in the beginning. Or was it just for demonstration purposes? @chrisma @alexanderernst

PR #137, opened 2016-12-04
17 daterange single day

tom95:
Forgot this acceptance criteria for #17, date ranges that encompass only a single day should also be displayed as a single date string. (As free extra boon this also i18ns the stuff [daterange display])


spec/features/event_spec.rb
```diff
@@ -70,6 +70,14 @@
       end
     end
 
+    it "should display a single day date range as a single date" do
+      event = FactoryGirl.create(:event, :single_day)
+      visit event_path(event)
+      expect(page).to have_text(event.date_ranges.first.start_date)
+      expect(page).to_not have_text(event.date_ranges.first.start_date.to_s +
+                                    ' bis ' + event.date_ranges.first.end_date.to_s)
```
bjrne: Shouldn't this use i18n as well?

tom95: Will be discussed during or after the next scrum of scrums meeting

Issue #18, opened 2016-11-10
US_1.4: Application deadline 🍪13


tobiduer:
**As** 
organizer
 
**I want to** 
enter an application deadline for events 

**in order to** 
close the application process automatically. 

**13 🍪**
- [ ] I can input a single day
- [ ] Application deadline needs to be before start of the event
- [ ] Show user that deadline is at midnight of picked date
- [ ] I can't choose dates in the past
- [ ] I get errormessages if  the case above happen
- [ ] It is clear to me which format I have to use to input my dates

PR #139, opened 2016-12-04
12 2.1 change application until deadline

nstrelow:


app/views/application_letters/_form.html.erb
```diff
@@ -84,7 +91,7 @@
 
   <div class="form-group">
     <div class="col-lg-offset-2 col-lg-10">
-      <%= f.submit nil, :class => 'btn btn-primary' %>
+      <%= f.submit nil, :class => 'btn btn-primary', :disabled => @application_letter.after_deadline? %>
```
cmfcmf: A user could simply remove the class and submit the form. A server-side check is needed!

cmfcmf (dismissed):
LGTM, except for server-side checking.

app/models/application_letter.rb
```diff
@@ -23,4 +23,12 @@ class ApplicationLetter < ActiveRecord::Base
 
   enum status: {accepted: 1, rejected: 0, pending: 2}
 
+  # Checks if the deadline is over
+  #
+  # @param none
+  # @return [Boolean] true if deadline is over
+  def after_deadline?
+    Date.today > event.application_deadline if event.present?
```
nstrelow: No, the same form is used for creating new applications which have no event

app/views/application_letters/_form.html.erb
nstrelow: Good point, will work on that

JuliusCosmoRomeo (approved):
seems to work properly (y)

app/models/application_letter.rb
```diff
@@ -20,7 +20,23 @@ class ApplicationLetter < ActiveRecord::Base
   validates :grade, presence: true, numericality: { only_integer: true }
   validates :vegeterian, :vegan, :allergic, inclusion: { in: [true, false] }
   validates :vegeterian, :vegan, :allergic, exclusion: { in: [nil] }
+  validate :deadline_cannot_be_in_the_past, :if => Proc.new { |letter| !(letter.status_changed?) }
```
cmfcmf: What is `Proc` and why is it needed? 
It isn't clear to me why you used it (after reading [the docs](https://ruby-doc.org/core-2.2.0/Proc.html).

cmfcmf (commented):
@nstrelow I would've liked to have a change to review the changes you made after I did my review. :)

ekrebs5:
I noticed something while testing:
You are using a deadline attribute here. But at the moment there is no way to edit it in the UI, thus it is always nil, thus some of your functions fail. Because of this I cannot create an application letter.

In short, creating an application letter is not possible on dev right now. Could this be fixed somehow?
( @Baschdl )

bjrne:
@ekrebs5 since this is already merged it could make sense to open an issue and link to this PR, this way the discussion of the problem probably moves to a more relevant place :) you could the either tag @Baschdl or all scrum masters if this depends on something another group built.

cmfcmf:
> You are using a deadline attribute here. But at the moment there is no way to edit it in the UI, thus it is always nil, thus some of your functions fail. Because of this I cannot create an application letter.

This is [part of a different user story](https://github.com/hpi-swt2/workshop-portal/issues/18) which is not handled by our team. The @hpi-swt2/scrum-master-2016 should already know about the problem.

I agree, however, that it's difficult to use at the moment.

nstrelow:
@ekrebs5 Thanks for noticing, I will check and fix issues I encounter.

nstrelow:
See #170 and #171

app/models/application_letter.rb
nstrelow: You need a Proc or lambda:
http://stackoverflow.com/questions/6232099/rails-validation-if-proc-new-or-lambda
http://stackoverflow.com/questions/8848601/rails-how-to-validate-a-field-only-if-a-another-field-has-a-certain-value

cmfcmf: I understand, thank you! I suppose a `lambda` would have been more "clear" for an inexperienced Rails developer like me. But anyways, good work!

PR #158, opened 2016-12-07
55 3.14 send emails backend

frederike-ramin:
Make it possible to send E-Mails via a simple method call.  
  
**IMPORTANT:** Have a look at the Wiki too, for more detailed information.


Baschdl:
I think it would be good to add the initial setup mentioned on the wiki page to the .travis.yml setup

aloeser:
@Baschdl I dont think this is a good idea/required. Figaro itself should be installed automatically via "bundle install".
The second and third line are for setting up email credentials for development and production mode. But since we dont want to send emails in tests, the tests are mocked and "deliver" the emails into an array. This is why we dont need the second and third command.
And the Heroku line is pointless without step 2 and 3, and should only executed once by someone with access rights to heroku, not with every deploy

Baschdl:
@tobiduer Do you/the POs want to try the mail functionality on heroku?

tobiduer:
@Baschdl: Yes. Please :)

frederike-ramin:
@Baschdl Why would you ask that? If course they want to try it..that's why a corresponding frontend is following by @TimBasel and @Lando-L.

Baschdl:
Then it would be great if the mail backend is automatically configured on heroku. Otherwise they can't test and show the mail functionality

aloeser:
@Baschdl As I said, this is nothing we want do in this Pull Request. We intentionally use the figaro gem to hide the email credentials. In order to get sending emails working on Heroku, someone with write access to Heroku needs to set the environment variable once. That is what I and @alexanderernst agreed on.

app/mailers/portal_mailer.rb
```diff
@@ -0,0 +1,13 @@
+class PortalMailer < ApplicationMailer
+
+  # @param hideRecipients [Boolean], recipients [Array<String>], reply_to [Array<String>], subject [String], content [String]
+  # @return [ActionMailer::MessageDelivery] a mail object with the given parameters.
+  def generic_email(hideRecipients, recipients, reply_to, subject, content)
```
querenker: I know, we haven't any conventions about this, but afaik it is very common in ruby to naming boolean flags with a question mark suffix. 
`hideRecipients -> hide_recipients?`

querenker (changes requested):

app/mailers/portal_mailer.rb
querenker: Thanks for the link :) I don't understand the reason nonetheless and for me it is even a violation of the uniform access principle. But probably there is a good reason for doing so :)

querenker (approved):

ekrebs5 (approved):

PR #168, opened 2016-12-09
Nicer date widget (fixes #155), switch Date.today for current (fixes #162)

### LINKED ISSUE #155: Well Formatted Error for Date Range Picker
- [ ] Bootstrap Date Range Picker verwenden, weil einheitlicher und schöner... https://github.com/Nerian/bootstrap-datepicker-rails
- [ ] error unter dem Date Range Picker anzeigen, nicht nur oben im Error "Header"
- [ ] Danach auch den error header ganz entfernen und durch das angelegte Partial ersetzen

- [ ] Das "X" verschwindet nachdem man ein Event mit mehreren Date Ranges und Fehlern versucht abzuschicken

### LINKED ISSUE #162: Replace Date.today with Date.current
Any occurence of Date.today should really be Date.current, because current is timezone-aware while today is not. See http://stackoverflow.com/questions/6635363/what-is-the-difference-between-date-current-and-date-today.

Quote: "you can have the situation where `Date.today == Date.tomorrow`" (becauseDate.tomorrow is timezone-aware...)

(Thanks for finding this @Blaidd-Drwg )

tom95:
This switches the date time ranges on event to use the bootstrap date picker. It also makes sure that the events page gets prettier error messages and that all date range related errors are correctly gathered for the event.

The datepicker is now enabled on all pages for all elements matching `[data-provide="datepicker"]`. Call enableDatepickers() if you need to enable new pickers at a time later than page load.

It also replaces all occurences of Date.today with Date.current.


app/views/layouts/application.html.erb
```diff
@@ -4,6 +4,15 @@
   <title>Workshop-Portal</title>
   <%= stylesheet_link_tag    'application', media: 'all', 'data-turbolinks-track' => true %>
   <%= javascript_include_tag 'application', 'data-turbolinks-track' => true %>
+  <script>
+    function enableDatepickers() {
+      $('[data-datepicker="true"]').datepicker({
+        language:"<%= I18n.locale %>",
+        format: "<%= I18n.t('date.formats.datepicker')%>",
```
cmfcmf: This should use the `escape_javascript` option: http://api.rubyonrails.org/classes/ActionView/Helpers/JavaScriptHelper.html#method-i-escape_javascript

PR #171, opened 2016-12-09
Check if event.application_deadline is not nil, Fixes #170

### LINKED ISSUE #170: Fix creation of new application letters without event.application_deadline #12

nstrelow:
We cannot create event application deadlines yet, so we cannot assure event.application_deadline is not nil.

This should never be nil, when #18 is finished. Please remove this in #18 .

### DIFF
--- a/app/models/application_letter.rb
+++ b/app/models/application_letter.rb
@@ -28,11 +28,12 @@ class ApplicationLetter < ActiveRecord::Base
   enum status: {accepted: 1, rejected: 0, pending: 2}
 
   # Checks if the deadline is over
+  # additionally only return if event and event.application_deadline is present
   #
   # @param none
   # @return [Boolean] true if deadline is over
   def after_deadline?
-    Date.today > event.application_deadline if event.present?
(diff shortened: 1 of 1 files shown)


app/models/application_letter.rb
```diff
@@ -28,11 +28,12 @@ class ApplicationLetter < ActiveRecord::Base
   enum status: {accepted: 1, rejected: 0, pending: 2}
 
   # Checks if the deadline is over
+  # additionally only return if event and event.application_deadline is present
```
cmfcmf: Please add a `TODO` statement.
```ruby
# TODO: 'event.application_deadline' should never be nil, when #18 is finished. Please remove this in #18.
```

cmfcmf (changes requested):

cmfcmf (approved):

nilsthamm (changes requested):
Please document your changes here:
https://github.com/hpi-swt2/workshop-portal/wiki/Application-Letter

nilsthamm (approved):

PR #174, opened 2016-12-09
Remove active column from events table. Closes #167

### LINKED ISSUE #167: Remove `active` field from event
`active` is now deprecated by `draft`.

bjrne:


app/controllers/events_controller.rb
```diff
@@ -94,6 +94,6 @@ def date_from_form(date_info)
 
     # Only allow a trusted parameter "white list" through.
     def event_params
-      params.require(:event).permit(:name, :description, :max_participants, :active, :kind, :organizer, :knowledge_level)
+      params.require(:event).permit(:name, :description, :max_participants, :draft, :kind, :organizer, :knowledge_level)
```
bjrne: @MasterCarl this should have been here in the first place, doesn't it? Or is it just not necessary for checkbox?

MasterCarl: Actually, `:draft` is contained directly in `params` since it depends on the button clicked and not on the form content. Therefore, I would leave this out

MasterCarl (changes requested):
:draft should not be part of params.require.permit

MasterCarl (approved):

PR #192, opened 2016-12-14
55 3.14 Send Email UI

TimBasel:


app/models/event.rb
```diff
@@ -101,13 +110,32 @@ def email_adresses_of_rejected_applicants
...
-    accepted_applications.map{ |application_letter| application_letter.user.email }.join(',')
+  # @return [Email] new acceptance email
+  def generate_acceptances_email
+    email = Email.new
+    email.hide_recipients = false
+    email.recipients = email_adresses_of_accepted_applicants
+    email.reply_to = 'workshop.portal@hpi.de'
+    email.subject = ''
+    email.content = ''
+    email
```
aloeser: Why do we need this line? Would it return '' otherwise? In that case a "return email" might be easier to understand?

app/views/emails/_email_form.erb
```diff
@@ -0,0 +1,45 @@
+<%= form_for @email, url: {action: 'send_email', controller: 'emails'}, html: {class: 'email-form form-horizontal'} do |f| %>
+    <div class="form-group">
+      <div class="col-sm-4 col-sm-offset-2 btn-group" data-toggle="buttons">
+        <label class="btn btn-default <%= 'active' if @email.hide_recipients %>">
+          <%= f.radio_button :hide_recipients, @email.hide_recipients, checked: @email.hide_recipients %>
+          <%= t('.to') %>
+        </label>
+        <label class="btn btn-default <%= 'active' if !@email.hide_recipients %>">
```
aloeser: I think we can use "unless" instead of "if !"

app/views/events/_applicants_overview.html.erb
```diff
@@ -63,51 +61,25 @@
...
-<!-- Bootstrap Modal for sending emails -->
-<div id="send-emails-modal" class="modal fade" role="dialog">
-  <div class="modal-dialog">
-
-    <!-- Modal content-->
-    <div class="modal-content">
-      <div class="modal-header">
-        <button type="button" class="close" data-dismiss="modal">&times;</button>
-        <h4 class="modal-title"></h4>
+    <% if (!@event.applications_classified?) %>
```
aloeser: maybe you could use an unless instead of the if ! ? I dont know if the elsif later on still works though.

config/locales/de.emails.yml
```diff
@@ -0,0 +1,14 @@
+# German translation for emails
+
+de:
+  emails:
+    send_email:
+      sending_successful: 'Email wurde erfolgreich versendet'
+    email_form:
+      to: 'An'
+      bcc: 'BCC'
```
aloeser: We dont need this one anymore (PO changed his opinion), so I suggest to remove it (and update the label on the buttons for (non) hiding the recipients)

spec/features/events_spec.rb
```diff
@@ -79,8 +79,25 @@
       FactoryGirl.create(:application_letter_accepted, :event => @event, :user => @pupil.user)
     end
     visit event_path(@event)
-    click_button I18n.t('events.applicants_overview.sending_acceptances')
-    expect(page).to have_selector('div', :id => 'send-emails-modal')
+    click_link I18n.t('events.applicants_overview.sending_acceptances')
+    fill_in('email_subject', with: 'Subject')
+    fill_in('email_content', with: 'Content')
+    click_button I18n.t('emails.email_form.send')
+  end
```
aloeser: Maybe I am just tired, but what exactly does this test do? It only tests if there is the UI interaction that is needed for sending emails, but whether emails are actually sent is not tested?

spec/features/events_spec.rb
```diff
@@ -79,8 +79,25 @@
...
+    2.times do |n|
+      @pupil = FactoryGirl.create(:profile)
+      @pupil.user.role = :pupil
+      FactoryGirl.create(:application_letter_rejected, :event => @event, :user => @pupil.user)
+    end
+    visit event_path(@event)
+    click_link I18n.t('events.applicants_overview.sending_rejections')
+    fill_in('email_subject', with: 'Subject')
+    fill_in('email_content', with: 'Content')
+    click_button I18n.t('emails.email_form.send')
```
aloeser: same as above

spec/views/emails/send_email.html.erb_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'rails_helper'
+
+RSpec.describe "emails/email_form", type: :view do
+  before(:each) do
+    @email = assign(:email, FactoryGirl.create(:email))
+    sign_in(FactoryGirl.create(:user, role: :admin))
```
aloeser: why admin?

app/views/emails/_email_form.erb
aloeser: I see what you mean. I think both options are okay

spec/features/events_spec.rb
TimBasel: yes, but is there a way to test if the Mailer.send_generic_email fails

aloeser: maybe you could test if an email is sent at all? you can find some inspiration in spec/services/mailer_spec.rb:
expect{<action that causes the mail to be sent>}.to change{ActionMailer::Base.deliveries.count}.by(1)
Note that hide_recipients needs to be false in that case, otherwise not 1 but 1 email for each recipient is sent

app/controllers/events_controller.rb
```diff
@@ -84,6 +84,22 @@ def participants
     @participants = @event.participants_by_agreement_letter
   end
 
+  # GET /events/1/send-acceptances-email
+  def send_acceptances_email
+    event = Event.find(params[:id])
+    @email = event.generate_acceptances_email
```
frederike-ramin: Why "acceptanceS"? isn't this singular?

app/models/email.rb
```diff
@@ -0,0 +1,19 @@
+class Email
+  include ActiveModel::Validations
+  include ActiveModel::Conversion
+  extend ActiveModel::Naming
+
+  attr_accessor :hide_recipients, :recipients, :reply_to, :subject, :content
+
+  validates_presence_of :hide_recipients, :recipients, :reply_to, :subject, :content
+
+  def initialize(attributes = {})
```
frederike-ramin: What kind of attributes? Mail attributes?

spec/features/events_spec.rb
```diff
@@ -79,8 +79,27 @@
       FactoryGirl.create(:application_letter_accepted, :event => @event, :user => @pupil.user)
     end
     visit event_path(@event)
-    click_button I18n.t('events.applicants_overview.sending_acceptances')
-    expect(page).to have_selector('div', :id => 'send-emails-modal')
+    click_link I18n.t('events.applicants_overview.sending_acceptances')
+    fill_in('email_subject', with: 'Subject')
+    fill_in('email_content', with: 'Content')
+    click_button I18n.t('emails.email_form.send')
+    expect(ActionMailer::Base.deliveries.last.to.count).to eq(1)
```
aloeser: with this line you only assure that the last email was adressed to exactly one person, but not that an email was sent

app/views/emails/_email_form.erb
```diff
@@ -0,0 +1,46 @@
...
+    </div>
+    <div class="form-group">
+      <label class="control-label col-sm-2" for="subject"><%= t('.subject') %>:</label>
+      <div class="col-sm-10">
+        <%= f.text_field :subject, class: 'email-subject form-control' %>
+      </div>
+    </div>
+    <div class="form-group">
+      <div class="col-sm-10 col-sm-offset-2">
+        <%= f.text_area :content, class: 'email-content', rows: 15 %>
```
frederike-ramin: How about a label for content?

app/views/events/_applicants_overview.html.erb
```diff
@@ -63,51 +61,25 @@
...
-<div id="send-emails-modal" class="modal fade" role="dialog">
-  <div class="modal-dialog">
-
-    <!-- Modal content-->
-    <div class="modal-content">
-      <div class="modal-header">
-        <button type="button" class="close" data-dismiss="modal">&times;</button>
-        <h4 class="modal-title"></h4>
+    <% if not @event.applications_classified? %>
+      <div class="btn-group pull-right tooltip-wrapper has-tooltip" role="group" data-toggle="tooltip" title="<%= t '.unclassified_applications_left'%>">
```
frederike-ramin: If you wanted to, you could make singular/plural of "Bewerbung(en)" dependent to how many applications this event has.

app/views/emails/_email_form.erb
```diff
@@ -0,0 +1,46 @@
+<%= form_for @email, url: {action: 'send_email', controller: 'emails'}, html: {class: 'email-form form-horizontal'} do |f| %>
+    <div class="form-group">
+      <label class="control-label col-sm-2" for="to"><%= t('.further_recipients') %>:</label>
+      <div class="col-sm-6 btn-group" data-toggle="buttons">
+        <label class="btn btn-default <%= 'active' if @email.hide_recipients %>">
+          <%= f.radio_button :hide_recipients, @email.hide_recipients, checked: @email.hide_recipients %>
+          <%= t('.show_recipients') %>
```
frederike-ramin: How about not naming it "An" and "Bcc" in the buttons instead hiding or not?

spec/controllers/emails_controller_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe EmailsController, type: :controller do
+
```
frederike-ramin: if nothing's in here, is it necessary?

spec/views/emails/send_email.html.erb_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+
+RSpec.describe "emails/email_form", type: :view do
+  before(:each) do
+    @email = assign(:email, FactoryGirl.build(:email))
+    sign_in(FactoryGirl.create(:user, role: :admin))
+    render :partial => "emails/email_form"
+  end
+
+  it("renders required email fields") do
+    expect(rendered).to have_field(t'emails.email_form.show_recipients')
```
frederike-ramin: What is this t? Is it translate? How can you call it without using I18n?

spec/views/emails/send_email.html.erb_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+    expect(rendered).to have_field(t'emails.email_form.hide_recipients')
+
+    expect(rendered).to have_field('email_recipients')
```
frederike-ramin: add reply_to field test.

app/views/emails/_email_form.erb
frederike-ramin: plus: add email-content form-control so it looks like the rest.

app/models/email.rb
TimBasel: general attributes, it implements the general interface for Email.new(params)
of cause in this case it only accepts email specific attributes, but the ActiveModel interface is general

spec/views/emails/send_email.html.erb_spec.rb
TimBasel: Inside View tests you can use the abbreviated version, but I will update to make it clearer

spec/views/emails/send_email.html.erb_spec.rb
frederike-ramin: @TimBasel Did you see that comment? Didn't fix it yet?

frederike-ramin: nvm..my fault.

PR #196, opened 2016-12-14
55 3.14 send emails backend avoid bcc

aloeser:
If recipients of an email should not know of each other, we now send multiple emails with to adressing instead of one with bcc. We do that in order to avoid a spam classification


TimBasel:
Wiki needs to be updated

TimBasel:
Wiki needs to be updated again

timbasel (approved):

ekrebs5 (approved):

Issue #100, opened 2016-11-30
US_4.1b: PDF is real PDF 💰 3


julisa99:
**As**
Admin

**I want to**
let pupil only submit real pdfs as letter of agreement

**in order to**
be safe from viruses.

**Acceptance criteria:**
Try to open the PDF on server-side. Throw an error, if it doesn't happen.
As pupil I can't upload a *.txt renamed in *.pdf .
### **TODO: Find and fix the bug that caused the Heroku server to crash**

PR #200, opened 2016-12-14
100 4.1b pdf is real pdf

h0rnyc0rny:
connects to #100


app/models/agreement_letter.rb
```diff
@@ -90,4 +95,13 @@ def too_big?(file)
     def wrong_filetype?(file)
       file.content_type != ALLOWED_MIMETYPE
     end
+
+  def wrong_file_trailer? file
+    begin
+    page_analysis = PDF::Inspector::Page.analyze_file(file.open)
```
Blaidd-Drwg: This variable is not needed.
Apparently the `.open` is not needed as well, at least the tests succeed without it.

app/models/agreement_letter.rb
```diff
@@ -90,4 +95,13 @@ def too_big?(file)
...
+    false
+    rescue
```
Blaidd-Drwg: Please only rescue `PDF::Reader::UnsupportedFeatureError` and `PDF::Reader::MalformedPDFError` to avoid hiding other exceptions.

app/models/agreement_letter.rb
```diff
@@ -90,4 +95,13 @@ def too_big?(file)
     def wrong_filetype?(file)
       file.content_type != ALLOWED_MIMETYPE
     end
+
+  def wrong_file_trailer? file
+    begin
+    page_analysis = PDF::Inspector::Page.analyze_file(file.open)
+    false
```
Blaidd-Drwg: Missing indentation

app/models/agreement_letter.rb
```diff
@@ -90,4 +95,13 @@ def too_big?(file)
     def wrong_filetype?(file)
       file.content_type != ALLOWED_MIMETYPE
     end
+
+  def wrong_file_trailer? file
```
Blaidd-Drwg: I don't think this function's name is descriptive of what it checks. Maybe `unable_to_open?`? Dunno, you'll probably have better ideas :wink:

Blaidd-Drwg:
Issue #100, Pull Request #200. I'm not sure what it means, but it sure means something :stuck_out_tongue_winking_eye:

PR #203, opened 2016-12-14
hotfix to allow accepting and rejecting applications

aloeser:
someone recently added "application_status_locked" as an attribute to the event model. From what i have seen, this value is not set on creating an event. This made it impossible to accept or reject applications, since the formular was only shown when application_status_locked is false, but it is nil or not set per default. Now it also shows on nil/not set


frederike-ramin:
Please add description why this hotfix is necessary, then I will approve.

timbasel (approved):

frederike-ramin:
Perhaps it is prettier to set the dafault to false then doing that workaround?

frederike-ramin:
Buuut then you would change the schema which we aren't supposed to..I see. Approved.

frederike-ramin (approved):

aloeser:
I could write a migration with an default value yes. But I dont really know the purpose of this field, so I would rather leave that task for the responsible team

PR #206, opened 2016-12-15
55 3.14 Fix Email Recipient Visibility & Copy Bug

TimBasel:


app/views/emails/_email_form.erb
```diff
@@ -2,12 +2,12 @@
...
-          <%= f.radio_button :show_recipients, !@email.hide_recipients, checked: !@email.hide_recipients %>
+        <label class="btn btn-default <%= 'active' if @email.hide_recipients %>">
+          <%= f.radio_button :hide_recipients, @email.hide_recipients, checked: @email.hide_recipients %>
           <%= t('.show_recipients') %>
         </label>
-        <label class="btn btn-default <%= 'active' if @email.hide_recipients %>">
-          <%= f.radio_button :hide_recipients, @email.hide_recipients, checked: @email.hide_recipients%>
+        <label class="btn btn-default <%= 'active' if !@email.hide_recipients %>">
+          <%= f.radio_button :hide_recipients, !@email.hide_recipients, checked: !@email.hide_recipients%>
           <%= t('.hide_recipients') %>
```
frederike-ramin: If I read this correctly, then "Hide recipients" is active if hide_recipients is false? That seems to be the other way around to me.

PR #221, opened 2016-12-16
202 1.20 fix date range errors. Fixes #202

### LINKED ISSUE #202: US_1.20: Error display of the date-range picker
**As** 
organizer
 
**I want to** 
see errors when they occur in multiple time spans only once but  with a red circle around every range that the error apllies too. 

**in order to** 
have an amazing overview

tom95:


bjrne (approved):
👍

Issue #115, opened 2016-12-02
US_1.16 Link not in ID 🍪3


Helene0:
**As** 
user

**I want to** 
be able to click the name/title on the index pages and not the ID of an item to go to the detail page

**in order to** 
have a better user experience
 
which pages?
Anfragen
Meine Bewerbungen
Benutzerverwaltung

** 🍪**
Acceptance Criteria
 - [ ] The ID is not visible anymore
 - [ ] I can click on the name or description and arrive and the detail page

bjrne:
@Helene0 Shouldn't the description be removed in #47? I guess it'll just be the name then?

bjrne:
not relevant anymore

PR #222, opened 2016-12-16
Remove Id column in requests#index and put link on topic instead

bjrne:
Still relates to #115, after discussion with our PO it became clear that the Id is not wanted in any table. This is the only one  I could find apart from events

### DIFF
--- a/app/views/requests/index.html.erb
+++ b/app/views/requests/index.html.erb
@@ -5,7 +5,6 @@
 <table class="table table-striped">
   <thead>
     <tr>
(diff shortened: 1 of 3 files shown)


Baschdl:
@bjrne You should maybe request some reviews directly

bjrne:
@Baschdl I would, but after discussing with our PO it turns out this is not enough for this user story. I will do this once this PR is ready :)

bjrne:
Outdated and not needed anymore

PR #224, opened 2016-12-25
124 3.21 age when the event starts

frederike-ramin:
Show the applicant's age when the event starts, not the current age. (On events details, application overview page and on generated pdf)

Take your time for reviewing and merry christmas^^

### DIFF
--- a/app/models/application_letter.rb
+++ b/app/models/application_letter.rb
@@ -83,4 +83,15 @@ class ApplicationLetter < ActiveRecord::Base
       errors.add(:event, "Die Bewerbungen wurden bereits bearbeitet, eine Statusänderung ist nicht mehr erlaubt.")
     end
   end
+
+  # Returns the age of the user based on the date the event starts
+  #
+  # @param none
+  # @return [Int] for age as number of years
+  def applicant_age_when_event_starts
+    start_date = event.start_date
+    birth_date = user.profile.birth_date
+    start_date_is_before_birthday = start_date.month > birth_date.month || (start_date.month == birth_date.month && start_date.day >= birth_date.day)
+    return start_date.year - birth_date.year - (start_date_is_before_birthday ? 0 : 1)
+  end
 end
--- a/app/models/application_letter.rb
+++ b/app/models/application_letter.rb
@@ -89,9 +89,6 @@ class ApplicationLetter < ActiveRecord::Base
   # @param none
   # @return [Int] for age as number of years
   def applicant_age_when_event_starts
-    start_date = event.start_date
-    birth_date = user.profile.birth_date
-    start_date_is_before_birthday = start_date.month > birth_date.month || (start_date.month == birth_date.month && start_date.day >= birth_date.day)
-    return start_date.year - birth_date.year - (start_date_is_before_birthday ? 0 : 1)
+    user.profile.age_at_time(event.start_date)
   end
 end
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -26,14 +26,13 @@ class Profile < ActiveRecord::Base
     self.birth_date >= 18.years.ago
   end
 
-  # Returns the age of the user based on the current date
+  # Returns the age of the user based on the given date
   #
   # @param none
   # @return [Int] for age as number of years
-  def age
-    now = Time.now.utc.to_date
-    current_time_is_before_birthday = now.month > birth_date.month || (now.month == birth_date.month && now.day >= birth_date.day)
-    return now.year - birth_date.year - (current_time_is_before_birthday ? 0 : 1)
+  def age_at_time (given_date)
+    given_date_is_before_birth_date = given_date.month > birth_date.month || (given_date.month == birth_date.month && given_date.day >= birth_date.day)
+    return given_date.year - birth_date.year - (given_date_is_before_birth_date ? 0 : 1)
   end
 
   # Returns the Full Name of the user
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -26,6 +26,14 @@ class Profile < ActiveRecord::Base
     self.birth_date >= 18.years.ago
   end
 
+  # Returns the age of the user based on the current date
+  #
+  # @param none
+  # @return [Int] for age as number of years
+  def age
+    return age_at_time(Time.now)
+  end
+
   # Returns the age of the user based on the given date
   #
   # @param none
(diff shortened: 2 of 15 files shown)


config/locales/de.events.yml
```diff
@@ -10,6 +10,7 @@ de:
       filter: "Filtern"
       status: "Status"
       details: "Details"
+      age_when_event_starts: "Alter <br> bei Beginn"
```
bjrne: I doubt this is the right time for formatting

app/models/application_letter.rb
```diff
@@ -83,4 +83,15 @@ def status_cannot_be_changed
...
+
+  # Returns the age of the user based on the date the event starts
+  #
+  # @param none
+  # @return [Int] for age as number of years
+  def applicant_age_when_event_starts
+    start_date = event.start_date
+    birth_date = user.profile.birth_date
+    start_date_is_before_birthday = start_date.month > birth_date.month || (start_date.month == birth_date.month && start_date.day >= birth_date.day)
+    return start_date.year - birth_date.year - (start_date_is_before_birthday ? 0 : 1)
```
bjrne: I don't have a better idea at the moment, but maybe someone knows how to improve the readability of these two lines

bjrne (commented):
Rest seems good 👍

app/models/application_letter.rb
querenker: This method is very similar to [User#older_than_required_age_at_start_date_of_event?](https://github.com/hpi-swt2/workshop-portal/blob/dev/app/models/user.rb#L80) and [Profile#age](https://github.com/hpi-swt2/workshop-portal/blob/dev/app/models/profile.rb#L33). I think it should be a good idea to provide this functionality in a module or something like this?

app/views/events/_applicants_overview.html.erb
```diff
@@ -34,13 +34,14 @@
     </thead>
     <tbody>
     <%
-      @application_letters.sort_by! {|l| l.user.profile.send(params[:sort]) } if params[:sort]
+      @application_letters.sort_by! {|l| l.user.profile.send(params[:sort]) } if params[:sort] && params[:sort] != 'applicant_age_when_event_starts'
+      @application_letters.sort_by! {|l| l.send(params[:sort]) } if params[:sort] && params[:sort] == 'applicant_age_when_event_starts'
       @application_letters.reverse! if params[:order] == 'descending'
       @application_letters.select {|l| l.user.profile }.each do | application_letter | %>
```
querenker: I know this was already there and is probably not part of this user story. But I think this code should not be in the view?

spec/models/application_letter_spec.rb
```diff
@@ -67,4 +67,12 @@
      application.status = :rejected
      expect(application).to be_valid
   end
+
+  it "calculates the correct age of applicant when event starts" do
```
querenker: There are some edge cases which are not covered in this test. One more reason to factor this functionality out :)

querenker (changes requested):
Otherwise it looks good to me :)

app/views/events/_applicants_overview.html.erb
frederike-ramin: Though, the sorting needs to be done on client-side. So the only option is view, right?

config/locales/de.events.yml
frederike-ramin: Yeah..I wasn't sure either^^
But I'm also not very sure if the fix is that much better..

app/views/events/_applicants_overview.html.erb
MasterCarl: I wrote this line, and I do think that it should probably go in the controller (although it's more of a visual aspect than business logic). Since view rendering is server-side, this could be in either place (controller or view).

app/views/events/_applicants_overview.html.erb
```diff
@@ -13,7 +13,7 @@
         <%= sort_caret(Profile.human_attribute_name(:gender), :gender) %>
       </th>
       <th>
-        <%= sort_caret(Profile.human_attribute_name(:age), :age) %>
+        <%= sort_caret(( t('.age') + " <br> " + t('.when_event_starts')), :applicant_age_when_event_starts ) %>
```
querenker: I think the problem is the <br> tag in general. The HTML should describe the structure of the document independent of the visual representation. There are some cases where a line break has semantic meaning like in adresses or poetries. But I think this isn't such a case and therefore the positioning should be handled with CSS files?
I don't know how to archive this with CSS myself, but probably there exists some CSS magic ;) For now I think it isn't that bad that it is on one line if the room is available. Otherwise it should break the line either way?

frederike-ramin: It doesn't break the line if you don't do it maually. Thing is, that it fits into the line, but that makes the columns besides that one smaller which leads to a linebreak in the status-update-toggle-fields. You know what I'm talking about?

app/models/profile.rb
```diff
@@ -26,14 +26,13 @@ def adult?()
     self.birth_date >= 18.years.ago
   end
 
-  # Returns the age of the user based on the current date
+  # Returns the age of the user based on the given date
   #
   # @param none
   # @return [Int] for age as number of years
-  def age
```
cmfcmf: I would've preferred to keep the `age` method:
```ruby
def age
   return age_at_time(Time.now)
end
```

app/views/events/_applicants_overview.html.erb
cmfcmf: A simple `style="max-width: 150px"` on the enclosing `<th>` can fix it :)

frederike-ramin: @cmfcmf You. Are. An. Angel. Thanks!

config/locales/de.events.yml
```diff
@@ -10,6 +10,7 @@ de:
       filter: "Filtern"
       status: "Status"
       details: "Details"
+      age_when_event_starts: "Alter bei Beginn"
```
querenker: There are two identity versions of this local. Maybe one can be replaced?

querenker (changes requested):

config/locales/de.events.yml
frederike-ramin: They have different contexts. It wouldn't make sense to join them together.

querenker (approved):

aloeser (approved):
works for me locally

ekrebs5 (approved):

PR #227, opened 2016-12-29
145 cons german requests page

radscheit:


app/views/requests/show.html.erb
```diff
@@ -1,10 +1,10 @@
 <%- model_class = Request -%>
 <div class="page-header">
-  <h1><%=t '.title', :default => model_class.model_name.human.titleize %></h1>
+  <h1><%=t '.title', :default => model_class.model_name.human(count: 2).titleize %></h1>
 </div>
 
 <dl class="dl-horizontal">
-  <dt><strong><%= model_class.human_attribute_name(:topics) %>:</strong></dt>
+  <dt><strong><%= t('activerecord.request.topics') %>:</strong></dt>
```
Blaidd-Drwg: What was the reason for changing this?

config/locales/de.default.yml
```diff
@@ -1,6 +1,9 @@
 ---
 de:
   activerecord:
+    request:
+        one: "Anfrage"
+        other: "Anfragen"
```
Blaidd-Drwg: This isn't needed, is it? The translation in duplicated in `config/locales/de.requests.yml`

Blaidd-Drwg (approved):

PR #235, opened 2017-01-04

ekrebs5:
If you press "send acceptance mails" or "send rejection mails" on the applicants overview, application status changing will be locked


app/models/event.rb
```diff
@@ -154,6 +154,17 @@ def compute_occupied_places
     application_letters.where(status: ApplicationLetter.statuses[:accepted]).count
   end
 
+  # locks the ability to change application statuses
+  def lock_application_status
+    application_status_locked = true
```
frederike-ramin: If you lock the status in db, why also lock it as a local variable if it's never used again?

spec/models/ability_spec.rb
```diff
@@ -200,4 +200,5 @@
     ability = Ability.new(user)
     expect(ability).to be_able_to(:manage, Request)
   end
+
```
frederike-ramin: remove please

spec/views/events/show.html.erb_spec.rb
```diff
@@ -62,4 +62,5 @@
     render
     expect(rendered).to have_link(t(:print_button_label, scope: 'events.badges'))
   end
+
```
frederike-ramin: guess what..^^

frederike-ramin (changes requested):

db/schema.rb
```diff
@@ -63,6 +63,16 @@
...
+  create_table "emails", force: :cascade do |t|
+    t.boolean  "hide_recipients"
+    t.string   "recipients"
+    t.string   "reply_to"
+    t.string   "subject"
+    t.string   "content"
+    t.datetime "created_at",      null: false
+    t.datetime "updated_at",      null: false
+  end
+
```
aloeser: Maybe it is just me, but why do we need a database table for emails? Considering the attributes I dont think its for drafts.

aloeser: And further. shouldnt  there be an migration for that too? (if it is needed)

aloeser (changes requested):

Gemfile.lock
```diff
@@ -335,8 +335,5 @@ DEPENDENCIES
   twitter-bootstrap-rails
   web-console (~> 2.0)
 
-RUBY VERSION
```
ekrebs5: This was changed automatically, it should be reverted now.

Gemfile.lock
```diff
@@ -336,7 +336,7 @@ DEPENDENCIES
   web-console (~> 2.0)
 
 RUBY VERSION
-   ruby 2.2.2p95
+  ruby 2.2.2p95
```
aloeser: whitespace missing

db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20161209180054) do
+ActiveRecord::Schema.define(version: 20161211123737) do
```
aloeser: You should revert the version change too, I think? Not 100% sure

db/schema.rb
ekrebs5: I think we added this when we tried something with emails (like email knows event and can thus lock it).
Anyway, since we now do something different, we definitely don't need this anymore. (The email model should probably be enough, but we can talk about this again when we have an issue concerning emails)

spec/features/events_spec.rb
```diff
@@ -145,6 +143,20 @@
     end
   end
 
+  scenario "logged in as Organizer I can lock the event application statuses by pressing one of the email buttons" do
+    login(:organizer)
+    @pupil = FactoryGirl.create(:profile)
+    @application_letter = FactoryGirl.create(:application_letter, event: @event, user: @pupil.user)
+    @application_letter.status = 'accepted'
```
aloeser: I think there was an :application_letter_accepted

aloeser (changes requested):

app/models/event.rb
```diff
@@ -154,6 +154,17 @@ def compute_occupied_places
...
+    update(application_status_locked: true)
+  end
+  # unlocks the ability to change application statuses
+  def unlock_application_status
```
ekrebs5: At the moment unlock is only used for testing.

db/schema.rb
ekrebs5: Since we changed nothing else this is now also reverted.

app/models/event.rb
aloeser: Maybe you could outsource it to the tests? I dont think having code only needed for tests in productive model code is good style

app/models/event.rb
ekrebs5: you are right, I think this is a remnant of something we tried when we had a bug, it is now removed

spec/features/events_spec.rb
```diff
@@ -145,6 +143,19 @@
     end
   end
 
+  scenario "logged in as Organizer I can lock the event application statuses by pressing one of the email buttons" do
+    login(:organizer)
+    @pupil = FactoryGirl.create(:profile)
+    @application_letter = FactoryGirl.create(:application_letter_accepted, event: @event, user: @pupil.user)
+    @application_letter.save
```
aloeser: If I'm not mistaken the save was only to save the new status, which is now already set in the factory

app/models/event.rb
```diff
@@ -154,6 +154,15 @@ def compute_occupied_places
     application_letters.where(status: ApplicationLetter.statuses[:accepted]).count
   end
 
+  # locks the ability to change application statuses
+  def lock_application_status
+    update(application_status_locked: true)
+  end
+  # unlocks the ability to change application statuses
+  def unlock_application_status
```
aloeser: I still think this does not belong into production code when its only used for testing

aloeser (changes requested):

app/models/event.rb
frederike-ramin: Although, Tobi requested to be extensible in the original issue so that the locking could be removed, if you'd like to transfer sb from "Nachrücker" to "Angenommen". In this case, that method would be helpful.

frederike-ramin (approved):

app/models/event.rb
aloeser: I see what you mean, but I'm not sure if our PO really had this in mind. Since when you unlock it again, everything can be changed (not only "Nachrücker durch Absagen"), Further this causes the send accept/reject mail buttons to appear again.

aloeser: And you should add tests for the model functions, even though they are simple

ekrebs5: Concerning unlock: Alex R and I will talk with our PO about this today. We will get a new user story today which has the new status system (Nachrücker etc), so our stuff probably will be changed again. (If unlock is superfluous to our issue we can remove it / move it to tests)

Send Mail buttons are always visible btw, hiding them was not asked in the user story (and would be not good at the moment, since locking happens on the Event Applicants Overview page, not the Mail page, so you could lock the event without yet sending the Mails or only one kind)

aloeser (approved):

PR #250, opened 2017-01-05
Add database fields requested in #217 to the requests model

corinnaj:
I also removed the existing fields "user_id" and "topics" and adjusted the tests. This introduces all the necessary fields to work with requests, so that all other user stories can be worked on.


bjrne (approved):
Looks fine to me, maybe a full-time developer can glance over this as well?

tom95:
Glances fine

Issue #60, opened 2016-11-23
US_5.3a: Sort Materials 


julisa99:
**As** 
organizer and coach

**I want to**
be able to provide materials

**in order to**
all participants get materials 

**Acceptance Criteria**
- There should be the possibility to add in the UI folders (eg. day1, group-blue, ...) to keep the material archive clear.

Baschdl:
@julisa99 This user story is difficult to understand for other members of the project. Could you briefly describe the material?

PR #251, opened 2017-01-05
36 5.3 upload materials

pinussilvestrus:
-  show a material area on the event-page
-  show a file upload button on the event-page
-  show every file in material-directory in material area
-  If the upload button is pressed, you can add any files
-  restrict view of material upload to organizer and coach


app/views/events/_material_area.html.erb
```diff
@@ -0,0 +1,35 @@
+<div class="container-fluid" id="material_area">
```
radscheit: Why fluid? We have a fixed width.

radscheit: What about a namespace for the id?

app/controllers/events_controller.rb
```diff
@@ -108,6 +109,26 @@ def send_rejection_emails
...
 
+  # POST /events/1/upload_material
+  def upload_material
+    event = Event.find(params[:event_id])
+    material_path = event.material_path
+    Dir.mkdir(material_path) unless File.exists?(material_path)
+
+    file = params[:file_upload]
+    unless is_file?(file)
+      redirect_to event_path(event), alert: t("events.material_area.no_file_given")
```
radscheit: Outside from views it is always a good idea to use the `I18n` namespace for `t()` methods.

radscheit (changes requested):

radscheit (dismissed):

app/views/events/_material_area.html.erb
```diff
@@ -0,0 +1,35 @@
+<div class="container" id="event_material_area">
```
radscheit: Checking this again, using `.container` is not the proper way using bootstrap since this template is already loaded in a `.container` block element.
bildschirmfoto 2017-01-05 um 21 40 36

Moreover it could be better using a block element like this for wrapping your table: `<div class="col-xs-12 col-sm-12 col-md-10 col-lg-10 col-xs-offset-0 col-sm-offset-0 col-md-offset-1 col-lg-offset-1"></div>`. This would result into a narrower table, since it only has two columns.

app/views/events/show.html.erb
```diff
@@ -17,6 +17,9 @@
   <%= render :partial => 'applicants_overview' %>
 <% end %>
 
+<%= render :partial => 'material_area' %>
+
```
radscheit: Some space (~30px) would be fine between your table and the buttons below.

radscheit (changes requested):
It seems strange to me that neither this story nor #60 is about to be able to download the files. Did I miss something?

radscheit (approved):

PR #253, opened 2017-01-05
Start page

tom95:
This branch redesigns the start page and events index page (list). It also globally enables the use of glyphicons, changes the global font and drops some of the styles generated by the scaffold to reduce conflicts with other bootstrap components. The locale files now contain several bits of "lorem ipsum" that will need to be replaced by the customer, as well as two example images.


app/views/application/index.html.erb
```diff
@@ -1 +1,39 @@
-<h1><%= t('hello') %></h1>
\ No newline at end of file
+<section class="front-section row">
+	<div class="container">
+		<%= image_tag "front-welcome.png", class: "pull-right img-rounded img-front-welcome" %>
```
radscheit: In case of responsive layouts `img-responsive` is in need, depending on your intention, see the introducing comment

app/views/application/index.html.erb
```diff
@@ -1 +1,39 @@
-<h1><%= t('hello') %></h1>
\ No newline at end of file
+<section class="front-section row">
+	<div class="container">
```
radscheit: Why you don't give the `container` class directly to the `<section>`?

app/views/application/index.html.erb
```diff
@@ -1 +1,39 @@
...
+		<%= image_tag "front-welcome.png", class: "pull-right img-rounded img-front-welcome" %>
+		<h1><%= t "start_page.hello" %></h1>
+		<p class="lead"><%= t "start_page.intro_text" %></p>
+	</div>
+</section>
+
+<section class="front-section row">
+	<div class="container">
+		<h2>Veranstaltungen
+			<a class="btn btn-primary" href="<%= events_path %>">
```
radscheit: More spacing between `<h2>` and `<a>`.

app/views/application/index.html.erb
```diff
@@ -1 +1,39 @@
...
+				<%= glyph 'calendar' %> <%= t "start_page.all_events" %>
+			</a>
+		</h2>
+		<p class="lead"><%= t "start_page.intro_text_events" %></p>
+
+		<div class="event-list col-md-10 col-md-offset-1">
```
radscheit: In case of responsive design it could be a good idea to add width and offset for smaller devices `col-sm-12 col-md-10 col-sm-offset-0 col-md-offset-1`.

app/views/events/_event.html.erb
```diff
@@ -0,0 +1,44 @@
...
+    </small>
+  </h3>
+
+  <p><%= truncate(event.description, length: Event::TRUNCATE_DESCRIPTION_TEXT_LENGTH) do
+    link_to " " + I18n.t('events.list.read_on'), event_path(event)
+  end %></p>
+
+  <%= event.application_deadline_label.try { |label| tag_label label, :default } %>
+  <%= event.duration_label.try { |label| tag_label label, :default } %>
+  <br>
```
radscheit: In my opinion it is always better to use margins instead of hardcoded line breaks especially when it comes to multi device styles.

app/views/layouts/application.html.erb
```diff
@@ -16,27 +16,40 @@
...
-  <%= nav_bar do %>
-    <%= menu_group do %>
-      <%= menu_items %>
-      <% if user_signed_in? %>
-        <%= drop_down current_user.name do %>
-          <%= dropdown_items %>
-        <% end %>
-      <% else %>
-        <%= menu_item t(:login, scope: 'navbar'), new_user_session_path %>
+<%= nav_bar brand: t(:app_name), responsive: true, static: :top do %>
```
radscheit: The `responsive` attribute makes no sense as long as you don't use responsive features at all.

radscheit (changes requested):
It doesn't use breakpoints breakpoint at all, but this was your intention maybe.

app/views/application/index.html.erb
tom95: .front-section gets the darker background color, so it needs to span full width

app/views/application/index.html.erb
tom95: col-sm-12 and col-sm-offset-0 is implied by default if no other info is given

app/views/layouts/application.html.erb
tom95: So far, I only loosely added support for responsiveness. This was one thing I found in the docs and went ahead to enable to save time later.

tom95:
@radscheit Thanks for the feedback! I addressed your issues and added basic support for responsiveness on the two pages.

app/views/application/index.html.erb
radscheit: Yip I see, maybe I've been nitpicky in this case. The true problem is outside of your story. The main layout loads the whole site content in an `container-fluid`, so that you need to use `row` to get rid of that padding. It is not wrong, but for me it is not the `bootstrap` way of doing things. But anyway i will approve this, it is just of matter of taste and the problem of the one who made the main layout.

radscheit (approved):

app/models/event.rb
```diff
@@ -154,6 +155,40 @@ def compute_occupied_places
     application_letters.where(status: ApplicationLetter.statuses[:accepted]).count
   end
 
+  # Returns a label listing the number of days to the deadline if
+  # it's <= 7 days to go. Otherwise returns nil.
+  #
+  # @return string containing the label or nil
+  def application_deadline_label
+    days = (application_deadline - Date.current).to_i
```
MasterCarl: When I visit /events, I get the error

> NoMethodError at /events
> undefined method `-' for nil:NilClass

bjrne: Where do you get this error? On your machine or on heroku?

MasterCarl: This was on my machine. After I dropped the database, the error went away

Issue #123, opened 2016-12-02
Application Review


tobiduer:
**As**
applicant

**I want to**
see the important information for my application before I confirm it

**in order to**
double check the information

# Tasks

# Acceptance criteria

- After submitting the application, the applicant sees his / her personal profile information and below the current application

- The age is replaced by the birthday

bjrne:
@tobiduer @hpi-swt2/scrum-master-2016 is this already contained in another user story?

tobiduer:
@bjrne: I do not think so. Did you find one?

bjrne:
@tobiduer still the case?

tobiduer:
has been done in #257

Baschdl:
@tobiduer @bjrne Please close issues not until PR is merged and approve by PO

tobiduer:
@Baschdl: This issue is a duplicate ;)

PR #257, opened 2017-01-06
185 3.23 application checking page

aloeser:
Add a application checking page, so participants can check their applications after creating/editing them


bjrne:
@tobiduer is this enough for #123 ?

tobiduer:
@bjrne: Yes. I closed the other issue.

querenker (changes requested):
I especially like the use of the different tests 👍

app/views/application_letters/check.html.erb
```diff
@@ -0,0 +1,64 @@
...
+    end
+    if @application_letter.vegan
+      habits.push(ApplicationLetter.human_attribute_name(:vegan))
+    end
+    if @application_letter.allergic
+      habits.push(ApplicationLetter.human_attribute_name(:allergic))
+    end
+  %>
+  <strong><%=t('.eating_habits') %>:</strong>
+  <%= habits.join(', ') %><br/>
```
aloeser: We did ask him, and he said this behaviour is exactly what is wanted

app/views/application_letters/check.html.erb
```diff
@@ -0,0 +1,64 @@
...
+    if @application_letter.vegeterian
+      habits.push(ApplicationLetter.human_attribute_name(:vegeterian))
+    end
+    if @application_letter.vegan
+      habits.push(ApplicationLetter.human_attribute_name(:vegan))
+    end
+    if @application_letter.allergic
+      habits.push(ApplicationLetter.human_attribute_name(:allergic))
+    end
+  %>
```
aloeser: So you would prefer to put do it in the controller and pass it as @habits? to the view?

querenker: Maybe a model method for the application_letter that returns the habits as a list?

aloeser:
@tobiduer I dont think that #123 is completely done with this story. #123 also requires to replace the age-Field by birthday, which is not (yet) part of this user story.

tobiduer:
@aloeser: OK than please show the birthday instead of the age.

ekrebs5 (approved):

timbasel (approved):

querenker (approved):

PR #259, opened 2017-01-06
118 5.5 download letters of agreement

Fabian-Sommer:


app/views/events/participants.html.erb
```diff
@@ -1,40 +1,66 @@
 <%- model_class = Event -%>
+<script type="text/javascript">
+  jQuery(function() {
+    $('#selectAll').click(function() {
```
radscheit: In enviroments with a lot of different scripts it is a good idea to avoid `$` as placehoder for `jQuery` or at least to define it on the functions stack. In some cases it can lead to problems.

app/views/events/participants.html.erb
```diff
@@ -1,40 +1,66 @@
...
+        </td>
+        <% if can? :print_agreement_letters, Event %>
+          <td>
+            <%= check_box_tag('selected_participants[]', participant.id) %>
+          </td>
         <% end %>
-      </td>
-    </tr>
+      </tr>
+      <% end %>
```
radscheit: Disable button when there is nothing to download avaiable.

app/views/events/participants.html.erb
```diff
@@ -1,40 +1,66 @@
...
+    </table>
+    <% if can? :print_agreement_letters, Event %>
+      <%= submit_tag t('events.agreement_letters_download.download_all_as'), name: "download_all" %>
+      <%= select_tag :download_type, options_for_select([['pdf'], ['zip']]) %>
     <% end %>
-  </table>
-  
+  <% end %>
```
radscheit: Please add some space between your buttons and the ones below.

app/views/events/participants.html.erb
```diff
@@ -1,40 +1,66 @@
...
+            <%= check_box_tag('selected_participants[]', participant.id) %>
+          </td>
         <% end %>
-      </td>
-    </tr>
+      </tr>
+      <% end %>
+    </table>
+    <% if can? :print_agreement_letters, Event %>
+      <%= submit_tag t('events.agreement_letters_download.download_all_as'), name: "download_all" %>
```
radscheit: Why not just 2 buttons:
+ All as PDF
+ All as Zip

That would reduce the number of needed clicks.

app/views/events/participants.html.erb
```diff
@@ -1,40 +1,66 @@
 <%- model_class = Event -%>
+<script type="text/javascript">
```
radscheit: Why do you use inline scripts? They have a save home in assets/javascripts.

app/views/events/participants.html.erb
radscheit: It would make sense to have that buttons on the right side, to show the connection to the checkboxes.

radscheit (changes requested):

radscheit:
@Fabian-Sommer It would be cool to have some sample data.

app/views/events/participants.html.erb
Fabian-Sommer: Putting them there creates a problem with turbolinks (code doesn't work when loading the page for the first time).

app/views/events/participants.html.erb
Fabian-Sommer: PO requested that there be only one button (I had two buttons implemented previously).

app/views/events/participants.html.erb
radscheit: Did you tried it with a Document.ready event?

radscheit: Moreover i would recommend to target specific checkboxes with a specific class to avoid side effects to later implementations.

tom95: I have found onclick to work best as compromise between turbolinks, testing with capybara and readability, e.g. like this: https://gist.github.com/tom95/d29c5488c5e0332f6d95f766a8bc7146

radscheit (approved):

PR #260, opened 2017-01-06
16 3.1 accept all button

querenker:
On the applicants overview (event index page) the accept_all button now has functionality.

Coming:
- only organizer sees button + tests
- button only visible if places are free + tests
- infinite participants + tests


spec/features/events_spec.rb
```diff
@@ -145,6 +145,19 @@
     end
   end
 
+  scenario "logged in as Organizer I can push the accept all button to accept all applicants" do
+    login(:organizer)
+    @event.max_participants = 3
+    2.times do |i|
+      @pupil = FactoryGirl.create(:profile)
+      @application_letter = FactoryGirl.create(:application_letter, event: @event, user: @pupil.user)
```
frederike-ramin: You're creating two applications of the same pupil, aren't you? That could lead to a problem once that isn't possible anymore.

frederike-ramin (changes requested):
The WIP is really important. :D

app/views/events/_applicants_overview.html.erb
```diff
@@ -53,11 +53,8 @@
 
   <div class="container-fluid no-padding">
     <div class="btn-group" role="group">
-      <%= link_to t('.accept_all'),
-          events_path, :class => 'btn btn-default'%>
-
-      <%= link_to t('.print_all'),
-                  print_applications_event_path(@event.id), :target => "_blank", :class => 'btn btn-default'%>
+      <%= link_to t('.accept_all'), event_accept_all_applicants_path(@event.id), :class => 'btn btn-default' %>
```
aloeser: Might this be an usecase for .to_param instead of .id?

spec/features/events_spec.rb
aloeser: I dont think they create two applications for of the same. I think the factory creates a new user instance every time a new profile is created, not sure tho.
But maybe there are better ways to create a user than to create a profile and take its user?

app/controllers/events_controller.rb
```diff
@@ -109,6 +109,15 @@ def send_rejection_emails
     render :email
   end
 
+  # GET /events/1/accept-all-applicants
+  def accept_all_applicants
```
aloeser, line 118: Please add a test for this function

config/routes.rb
```diff
@@ -30,6 +30,7 @@
   get 'events/:id/participants' => 'events#participants'
   get 'events/:id/send-acceptance-emails' => 'events#send_acceptance_emails', as: :event_send_acceptance_emails
   get 'events/:id/send-rejection-emails' => 'events#send_rejection_emails', as: :event_send_rejection_emails
+  get 'events/:id/accept_all_applicants' => 'events#accept_all_applicants', as: :event_accept_all_applicants
```
aloeser, line 37: Please add a routing test for this line

aloeser (changes requested):

spec/features/events_spec.rb
```diff
@@ -147,11 +147,8 @@
 
   scenario "logged in as Organizer I can push the accept all button to accept all applicants" do
     login(:organizer)
-    @event.max_participants = 3
-    2.times do |i|
-      @pupil = FactoryGirl.create(:profile)
-      @application_letter = FactoryGirl.create(:application_letter, event: @event, user: @pupil.user)
-    end
+    @event = FactoryGirl.create :event, :with_diverse_open_applications
+    #@event.max_participants = 3
```
bjrne: Is this comment necessary or obsolete?

spec/features/events_spec.rb
```diff
@@ -144,6 +144,16 @@
     end
   end
 
+  scenario "logged in as Organizer I can push the accept all button to accept all applicants" do
+    login(:organizer)
+    @event = FactoryGirl.create :event, :with_diverse_open_applications
+    #@event.max_participants = 3
```
bjrne: Is this comment necessary or obsolete?

spec/models/event_spec.rb
```diff
@@ -180,6 +180,14 @@
     expect(email).to have_attributes(hide_recipients: false, recipients: event.email_adresses_of_rejected_applicants, reply_to: 'workshop.portal@hpi.de', subject: '', content: '')
   end
 
+  it "accepts all its application letters" do
+    event = FactoryGirl.create :event, :with_diverse_open_applications
+    #@event.max_participants = 3
```
bjrne: same here?

spec/features/events_spec.rb
ekrebs5: it will be later relevant to the user story, but it is not relevant to this pr (since max_participants is not relevant yet)
i changed the comment to be more precise

spec/models/event_spec.rb
ekrebs5: This is not necessary, since the model function should always work, no matter what max_participants is. I removed it

db/schema.rb
```diff
@@ -67,14 +67,15 @@
     t.string   "name"
     t.string   "description"
     t.integer  "max_participants"
-    t.datetime "created_at",                            null: false
-    t.datetime "updated_at",                            null: false
-    t.integer  "kind",                      default: 0
+    t.datetime "created_at",                             null: false
+    t.datetime "updated_at",                             null: false
+    t.integer  "kind",                       default: 0
```
aloeser: Please remove those whitespaces

spec/controllers/events_controller_spec.rb
```diff
@@ -170,6 +170,14 @@
         expect(assigns(:email)).to have_attributes(hide_recipients: false, recipients: @event.email_adresses_of_rejected_applicants, reply_to: 'workshop.portal@hpi.de', subject: '', content: '')
       end
     end
+
+    describe "GET #accept_all_applicants" do
+      it "should accept all application letters" do
```
aloeser: I dont think this is something you should test in a controller test. That belongs into model & feature tests.
Instead it should check whether the variables are set, if it renders the correct template/redirects correctly, etc.

app/views/events/_applicants_overview.html.erb
ekrebs5: What would this change? We only need the id?

app/views/events/_form.html.erb
```diff
@@ -33,6 +33,8 @@
     <div class="col-lg-10">
       <%= f.text_field :max_participants, :class => 'form-control' %>
       <%=f.error_span(:max_participants) %>
+      <%= f.label :participants_are_unlimited %>
```
aloeser, line 36: Do we really need an extra (database) field? Cant we compute it from the max_participants value?

aloeser (changes requested):

spec/controllers/events_controller_spec.rb
ekrebs5: it now tests the redirect to itself

timbasel: than the test should be renamed

app/views/events/_form.html.erb
querenker: The max_participants value is an integer that has to be greater than 0 for a valid event. As far as I know Ruby Integers (practically) has no limit. So I can't see any (magic) values that can be used to store that information?

app/views/events/_applicants_overview.html.erb
querenker: As far as I know in the default case it doesn't make any different, because `to_param` will call `id`. But it is possible to override the `to_param` method to provide more readable pathes for example `www.example.com/events/1` could be changed to `www.example.com/events/mint_camp_2016`. Currently it is inconsistent in general so that changes like this will cause in a lot of changes either way.

app/views/events/_form.html.erb
aloeser: You could use 0, since 0 does not have a meaning yet. I'm not really happy with either of the solutions

querenker: Unfortunately 0 is not grater than 0 and I would not change the validations for that hack. I think an extra (boolean) value is not that bad and normally it will be encapsulated through the getter and setter for max_participant.

app/views/events/_applicants_overview.html.erb
```diff
@@ -1,6 +1,10 @@
 <div class="container-fluid" id="applicants_overview">
   <h3><%= t '.title', title: @event.name %></h3>
-  <div id="free_places"><%= t '.free_places', count: @free_places %></div>
+  <%if @free_places == Float::INFINITY %>
```
aloeser: please add whitespaces after <% (and after the <%else too)

app/views/events/_applicants_overview.html.erb
aloeser: Then we both have he same understanding of it, and to_param seems the better/more flexible way to me.
But you are right that its pretty inconsistent, so this is maybe something the scrum masters/all teams should agree on? @Baschdl

app/views/events/_form.html.erb
aloeser: I see your point. None of the solutions sounds nice, so we can keep this as well, I guess

spec/views/events/show.html.erb_spec.rb
```diff
@@ -64,6 +64,26 @@
     expect(rendered).to have_link(t(:print_button_label, scope: 'events.badges'))
   end
 
+  it "should not display accept-all-button for non-organizers" do
+    sign_in(FactoryGirl.create(:user, role: :coach))
```
aloeser: what about pupils/participants?

spec/views/events/show.html.erb_spec.rb
```diff
@@ -64,6 +64,26 @@
...
+  it "should display accept-all-button for organizers if there are enough free places" do
+    sign_in(FactoryGirl.create(:user, role: :organizer))
+    @event.max_participants = Float::INFINITY
+    render
+    expect(rendered).to have_link(I18n.t('events.applicants_overview.accept_all'))
+  end
+
+  it "should not display accept-all-button if there are not enough free places" do
+    sign_in(FactoryGirl.create(:user, role: :organizer))
+    @event.max_participants = 0
```
aloeser: This is probably not the best test data, since @querenker mentioned in another discussion that events with 0 max_participants are not valid

aloeser (changes requested):
looks pretty good, only some small changes, and I need to test it locally tomorrow

app/models/event.rb
```diff
@@ -22,6 +22,25 @@ class Event < ActiveRecord::Base
   has_many :date_ranges
   accepts_nested_attributes_for :date_ranges
 
+  # Setter for max_participants
+  # @param [Int Float] the max number of participants for the event or infinity if it is not limited
+  # @return none
+  def max_participants=(value)
```
frederike-ramin, line 28: is that some kind of idiom to name a method like that?

app/models/event.rb
```diff
@@ -22,6 +22,25 @@ class Event < ActiveRecord::Base
...
+    if value == Float::INFINITY
+      self[:participants_are_unlimited] = true
```
frederike-ramin, line 30: and the max:participants value stays empty? Why not setting it to infinity?

app/models/event.rb
```diff
@@ -22,6 +22,25 @@ class Event < ActiveRecord::Base
...
+      self[:participants_are_unlimited] = false
+      self[:max_participants] = value
+    end
+  end
+
+  # Getter for max_participants
+  # @param none
+  # @return [Int Float] the max number of participants for the event or infinity if it is not limited
+  def max_participants
+    participants_are_unlimited ? Float::INFINITY : self[:max_participants]
```
frederike-ramin, line 41: Aaare you calling the same method here? Isn't that an infinite loop? (In assumption, that self[:max_participants] calls the max_participants method, which would be that one right here.)

app/views/events/_applicants_overview.html.erb
```diff
@@ -1,6 +1,10 @@
 <div class="container-fluid" id="applicants_overview">
   <h3><%= t '.title', title: @event.name %></h3>
-  <div id="free_places"><%= t '.free_places', count: @free_places %></div>
+  <% if @free_places == Float::INFINITY %>
```
frederike-ramin, line 3: if you even check for infinity here, why not also setting max_participants to infinity if it is unlimited?!

spec/controllers/events_controller_spec.rb
```diff
@@ -170,6 +170,13 @@
         expect(assigns(:email)).to have_attributes(hide_recipients: false, recipients: @event.email_adresses_of_rejected_applicants, reply_to: 'workshop.portal@hpi.de', subject: '', content: '')
       end
     end
+
+    describe "GET #accept_all_applicants" do
+      it "should redirect to the event" do
+        get :accept_all_applicants, id: @event.to_param, session: valid_session
+        expect(response).to redirect_to(@event)
```
frederike-ramin, line 177: That's not all it should do, right? There is plenty more functionality.

spec/features/events_spec.rb
```diff
@@ -146,17 +146,83 @@
...
     visit event_path(@event)
     click_link I18n.t "events.applicants_overview.accept_all"
     application_letters = ApplicationLetter.where(event: @event.id)
     expect(application_letters.all? { |application_letter| application_letter.status == 'accepted' }).to eq(true)
   end
 
+  scenario "logged in as Organizer and viewing the participants page all checkboxes are checked when pressing the \"check all\" button", js: true do
+    login(:organizer)
+    @user = FactoryGirl.create(:user)
+    @profile = FactoryGirl.create(:profile, user: @user, birth_date: 15.years.ago)
```
frederike-ramin, line 159: Why would you make the user underaged? Wouldn't it be easier and more understandable if the user was an adult, thus no agreement letter is needed?

spec/features/events_spec.rb
```diff
@@ -146,17 +146,83 @@
...
+    @application = FactoryGirl.create(:application_letter_accepted, user: @user, event: @event)
+    @agreement = FactoryGirl.create(:agreement_letter, user: @user, event: @event)
+    visit event_participants_path(@event)
+    check 'select_all_participants'
+    all('input[type=checkbox]').each do |checkbox|
+      expect(checkbox).to be_checked
+    end
+  end
+
+  scenario "logged in as Organizer when I want to download agreement letters but no participants are selected, it displays error message" do
```
frederike-ramin, line 170: This whole agreement stuff wasn't part of this story, was it?

frederike-ramin (changes requested):

spec/features/events_spec.rb
ekrebs5: this is probably from dev, we did nothing with agreement letters

spec/controllers/events_controller_spec.rb
ekrebs5: I think it is all that should be testet here?

What it does: it calls a function (event accept all...) and redirects
And the function functionality is tested elsewhere

spec/features/events_spec.rb
ekrebs5: this is probably from dev as well, we did nothing with agreement letters

app/views/events/_applicants_overview.html.erb
ekrebs5: I'm not sure what you mean?
The unlimited flag is true, thus max_participants is infinity (in theory, in practice integers can't be floats and integers don't support infinity standard-ly), thus free_places returns infinity

(this is also mostly a workaround until we know how to translate "infinity" automatically)

spec/features/events_spec.rb
frederike-ramin: Sorry, my bad.

app/views/events/_applicants_overview.html.erb
frederike-ramin: But why creating that flag at all if you can just set max:participants to infinity?

app/models/event.rb
ekrebs5: Because max_participants is an integer
But integers don't automatically support infinity, since they have no values that could be used (floats have values like 1/0 that will probably not be used for anything, those are then used for stuff like infinity)

app/models/event.rb
ekrebs5: I think it is the default way to name setters (in rails)

spec/views/events/show.html.erb_spec.rb
ekrebs5: They don't see it because they don't see the applicants overview in general, but it probably would be good to have it in the test explicitly, so I added it

spec/views/events/show.html.erb_spec.rb
ekrebs5: good point, I changed it

app/models/event.rb
frederike-ramin: Just change it to a float then. ;)

app/models/event.rb
frederike-ramin: @querenker @ekrebs5

spec/views/events/show.html.erb_spec.rb
```diff
@@ -64,6 +64,33 @@
...
+    @event.max_participants = Float::INFINITY
+    render
+    expect(rendered).to have_link(I18n.t('events.applicants_overview.accept_all'))
+  end
+
+  it "should not display accept-all-button if there are not enough free places" do
+    sign_in(FactoryGirl.create(:user, role: :organizer))
+    @event.max_participants = 1
+    2.times do
+      @application_letter = FactoryGirl.create(:application_letter, user: FactoryGirl.create(:user, role: :admin), event: @event)
```
aloeser: Is there a reason to explicitly set the role (to admin?)

db/migrate/20170110132223_add_unlimited_participants_to_events.rb
```diff
@@ -0,0 +1,5 @@
+class AddUnlimitedParticipantsToEvents < ActiveRecord::Migration
+  def change
+    add_column :events, :participants_are_unlimited, :boolean
```
aloeser: what happens with existing events? I think the default value will be nil, so its neither true nor false?
The way its implemented in the model atm will return the correct value, but maybe you could add an default value (false) anyway?

spec/views/events/show.html.erb_spec.rb
```diff
@@ -64,6 +64,33 @@
     expect(rendered).to have_link(t(:print_button_label, scope: 'events.badges'))
   end
 
+  it "should not display accept-all-button for non-organizers" do
+    [:coach, :student].each do | each |
+      sign_in(FactoryGirl.create(:user, role: each))
+      render
```
aloeser, line 71: Dont you need to set the max participants to unlimited before this line? Otherwise this test could be green just because there are not enough places

app/views/events/_applicants_overview.html.erb
```diff
@@ -53,11 +57,10 @@
 
   <div class="container-fluid no-padding">
     <div class="btn-group" role="group">
-      <%= link_to t('.accept_all'),
-          events_path, :class => 'btn btn-default'%>
-
-      <%= link_to t('.print_all'),
-                  print_applications_event_path(@event.id), :target => "_blank", :class => 'btn btn-default'%>
+      <% if (can? :edit_applicants, Event) && (@event.application_letters.count <= @event.max_participants) %>
```
aloeser, line 60: Maybe this could be moved to the controller, not sure tho

aloeser (changes requested):

app/models/event.rb
aloeser: self[:max_participants] directly accesses the model attribute. It would be an infinite loop, if they would call max_participants (which is usally an autogenerated simple getter, but here overridden)

app/views/events/_applicants_overview.html.erb
querenker: If I got you right you would like to have an variable `can_accept_all` or something like that which is set by the controller? I am not sure if this is a good idea. As view partial it could be also used in other event pages?

aloeser (approved):

frederike-ramin (approved):

db/migrate/20170110132223_add_unlimited_participants_to_events.rb
```diff
@@ -0,0 +1,5 @@
+class AddUnlimitedParticipantsToEvents < ActiveRecord::Migration
+  def change
+    add_column :events, :participants_are_unlimited, :boolean, default: false
```
bjrne, line 3: This leads to 
` t.boolean  "participants_are_unlimited", default: false`
im my schema.rb file when migrating. But here, the `, default: false` is missing in the `schema.rb`. Is there a reason why? @Baschdl @querenker @aloeser @frederike-ramin

aloeser: @bjrne The default value was added recently during the PR, so I guess the migration wasnt run again after that change, and no one of us kept in mind that the `schema.rb` needs to be updated too.
TL;DR default value should be there, we will fix this soon

Issue #28, opened 2016-11-10
US_4.5: Eating habits & allergies 💰 3


tobiduer:
**As**
organizer 

**I would like to**
see a list of allergies and eating habits of the participants in the event that I organize 

**in order to**
discuss this with the food provider. 

**Acceptance Criteria**
- As organizer I can see on the event page the participants of the event and their allergies/eating information.
- The organizer can:
  - sort by eating habits groups (vegan, vegetarian and none)
  - or show allergies at the top
  - show detail allergies information: While holding the mouse on the symbol, you can see more information.
  - option for printing/generate pdf
  - show vegan, vegetarian and allergies symbol all the time (none have no symbol, and the following symbols are from iconfinder with cc0-license)
![vegan](https://cloud.githubusercontent.com/assets/8758907/20611166/49ce1042-b2a5-11e6-8234-97174879c6d7.png)
![allergien](https://cloud.githubusercontent.com/assets/8758907/20611164/49cac1da-b2a5-11e6-8579-e72d74230827.png)
![cheese](https://cloud.githubusercontent.com/assets/8758907/20611167/49cf4e12-b2a5-11e6-9937-9c471224b8fb.png)

<img width="635" alt="eating_allergies" src="https://cloud.githubusercontent.com/assets/8758907/20611192/e1fe3702-b2a5-11e6-86c9-5628569dffe2.PNG">

See [UI wireframe for details](https://wireframe.cc/R3GJ9g)

tobiduer:
Dependency: Event, Role Pupil, Role Organizer

radscheit:
@julisa99 Please provide the symbols mentioned in the user story :)

radscheit:
@julisa99 Are the dependencies missing here?

radscheit:
@julisa99 @tobiduer Is this US still up to date?

julisa99:
Ja, wie kommst du darauf, dass sich das geändert hat?

bjrne:
@pinussilvestrus Is this a duplicate? Please state why you are closing issues that have no branch/PR history or duplicates linked.

PR #262, opened 2017-01-07
28 4.5 eating habits allergies

radscheit:
(#28)


spec/views/events/participants.html.erb_spec.rb
```diff
@@ -48,5 +48,83 @@
...
+
+  #     @user6 = FactoryGirl.create(:user, name: 'abc')
+  #     @profile6 = FactoryGirl.create(:profile, user: @user6, birth_date: 16.years.ago)
+  #     @application6 = FactoryGirl.create(:application_letter_accepted, user: @user6, event: @event)
+  #     @agreement6 = FactoryGirl.create(:agreement_letter, user: @user6, event: @event)
+      
+
+  #     render.extra_params()
+  #     expect(rendered).to eq([@user2, @user4, @user6, @user1, @user5, @user3])
+  #   end
```
bjrne: What's the purpose of all this comment-code? Since we have git, I'd prefer to delete this for the moment if it is not a mistake to have it as a comment here.

radscheit: Yes you are right, the next commit ist on the way...

lib/pdf_generation/participants_pdf.rb
```diff
@@ -0,0 +1,53 @@
...
+      text @event.date_ranges[0].to_s, size: 10
+      move_down 20
+      text summary, size: 10
+      move_down 20
+      #text @participants_allergic_application_letters.inspect
+      @participants_allergic_application_letters.each do |p|
+        output = p.user.name  
+        output += " (" + t('activerecord.attributes.application_letter.vegan') + ")" if p.vegan
+        output += " (" + t('activerecord.attributes.application_letter.vegeterian') + ")" if p.vegeterian
+        output += " <> " + p.allergies.to_s
```
Fabian-Sommer: Why is "<>" used here? ":" seems more intuitive.

app/views/events/participants.html.erb
```diff
@@ -7,44 +7,94 @@
...
+             </li>
+           <% end %>                
+           <% if application_letter.vegeterian %>
+             <li>
+               <img 
+                 class="img img-responsive eating-habit--icon"
+                 src="<%= image_url('application_letter_vegeterian.png') %>"
+               >
+             </li>
+           <% end %>
```
Fabian-Sommer: I am assuming that all vegans are also vegetarians. Showing both symbols when a person is both vegan and vegetarian is confusing.

Fabian-Sommer (dismissed):

lib/pdf_generation/participants_pdf.rb
radscheit: Ok, i will update this.

app/views/events/participants.html.erb
radscheit: Yes it is, but the logic of this model as well as the well as the validation of user input is not on the side of my user story. I can follow your argument but that is to much interpretation adn it would be part of another user story.

app/helpers/applicants_overview_helper.rb
```diff
@@ -11,8 +11,9 @@ def sort_caret(label, attr)
   end
 
   def sort_application_letters
-    @application_letters.sort_by! {|l| l.user.profile.send(params[:sort]) } if params[:sort] && params[:sort] != 'applicant_age_when_event_starts'
-    @application_letters.sort_by! {|l| l.send(params[:sort]) } if params[:sort] && params[:sort] == 'applicant_age_when_event_starts'
+    @application_letters.sort_by! {|l| l.user.profile.send(params[:sort]) } if params[:sort] && params[:sort] != 'applicant_age_when_event_starts' && params[:sort] != 'eating-habits'
+    @application_letters.sort_by! {|l| l.send(params[:sort]) } if params[:sort] && params[:sort] == 'applicant_age_when_event_starts' && params[:sort] != 'eating-habits'
+    @application_letters.sort! {|a,b| a.get_eating_habit_state <=> b.get_eating_habit_state } if params[:sort] && params[:sort] == 'eating-habits'
```
Blaidd-Drwg: These lines are pretty long, this would benefit from some refactoring.

app/helpers/participants_overview_helper.rb
```diff
@@ -0,0 +1,10 @@
+module ParticipantsOverviewHelper
+
+  def sort_participants_application_letters
+    @participants.sort! {|a,b| 
+    	a = ApplicationLetter.where(user_id: a.id, event_id: @event.id).first 
+    	b = ApplicationLetter.where(user_id: b.id, event_id: @event.id).first
+      a.get_eating_habit_state <=> b.get_eating_habit_state } if params[:sort] && params[:sort] == 'eating-habits'
```
Blaidd-Drwg: Usually you would use `{ block }` if it fits in a single line and `do block end` if the block takes multiple lines.
Also, putting the `if` after the statement is confusing if the statement spans multiple lines, the traditional `if conditional statement end` would be more readable.

app/helpers/participants_overview_helper.rb
```diff
@@ -0,0 +1,10 @@
+module ParticipantsOverviewHelper
+
+  def sort_participants_application_letters
+    @participants.sort! {|a,b| 
+    	a = ApplicationLetter.where(user_id: a.id, event_id: @event.id).first 
+    	b = ApplicationLetter.where(user_id: b.id, event_id: @event.id).first
```
Blaidd-Drwg: `ApplicationLetter.find_by(user_id: a.id, event_id: @event.id)` better describes what these lines do

app/views/events/participants.html.erb
```diff
@@ -7,44 +7,94 @@
...
+     <% if can? :show_eatings_habits, Event %>
+      <th>
+        <%= sort_caret(t('activerecord.methods.application_letter.eating_habits'), 'eating-habits') %>
+      </th>
+      <% end %>    
+    </tr>  
+  
+    <%  sort_participants_application_letters 
+    @participants.each do |participant| 
+       application_letter = ApplicationLetter.where(user_id: participant.id, event_id: @event.id).first
```
Blaidd-Drwg: You can use `find_by` here as well

app/views/events/participants.html.erb
```diff
@@ -7,44 +7,94 @@
...
+      </th>
+     <% end %>
+     <% if can? :show_eatings_habits, Event %>
+      <th>
+        <%= sort_caret(t('activerecord.methods.application_letter.eating_habits'), 'eating-habits') %>
+      </th>
+      <% end %>    
+    </tr>  
+  
+    <%  sort_participants_application_letters
```
Blaidd-Drwg: The sorting should be done in the controller.

app/views/events/participants.html.erb
```diff
@@ -7,44 +7,94 @@
...
+           <% if application_letter.allergic %>
+             <li>
+               <img 
+                 class="img img-responsive eating-habit--icon has-popover"
+                 src="<%= image_url('application_letter_allergic.png') %>"
+                 data-toggle="popover" 
+                 tabindex="50" 
+                 data-content="<%= application_letter.allergies %>" 
+                 data-placement="bottom"
+               >
```
Blaidd-Drwg: The image_tag helper is a touch more readable (same for the other images)
```ruby
               <%= image_tag('application_letter_allergic.png',
                             class: "img img-responsive eating-habit--icon has-popover",
                             tabindex: 50,
                             data: { toggle: "popover", content: application_letter.allergies, placement: "bottom" }) %>
```

app/views/profiles/show.html.erb
```diff
@@ -17,6 +17,13 @@
 <%= link_to t('.edit', :default => t("helpers.links.edit")),
               edit_profile_path(@profile), :class => 'btn btn-default' %>
 
+<%= link_to t('.destroy', :default => t("helpers.links.destroy")),
+              profile_path(@profile),
+              :method => 'delete',
+              :data => { :confirm => t('.confirm', :default => t("helpers.links.confirm", :default => 'Are you sure?')) },
+              :class => 'btn btn-danger' %>
```
Blaidd-Drwg: Was this required? It proabably shouldn't be allowed to delete your profile...

db/sample_data/application_letters.rb
```diff
@@ -66,21 +66,37 @@ def application_letter_applicant_programmierkurs_1(user, event)
...
     grade: 10,
     experience: "Suche im Internet",
     coding_skills: "keine",
     emergency_number: "01234567891",
     vegeterian: true,
     vegan: false,
-    allergic: false,
-    allergies: "",
+    allergic: true,
+    allergies: "WIMF",
```
Blaidd-Drwg: Obvious typo, you probably meant "Katy Perry? Echt jetzt, Jungs?"

lib/pdf_generation/participants_pdf.rb
```diff
@@ -0,0 +1,87 @@
...
+    @event = event
+    @document = Prawn::Document.new(page_size: 'A4')
+    @participants_count = @event.participants.count
+    @participants_allergic_application_letters = []
+    @participants_vegan_application_letters = []
+    @participants_vegeterian_application_letters = []
+    @participants_omnivorous_application_letters = []
+    @event.participants.each do |p|
+
+      letter = ApplicationLetter.where(user_id: p.id, event_id: @event.id).first
```
Blaidd-Drwg: `find_by`

lib/pdf_generation/participants_pdf.rb
```diff
@@ -0,0 +1,87 @@
...
+      if !letter.nil?
+        @participants_allergic_application_letters.push(letter) if letter && letter.allergic
+        @participants_vegeterian_application_letters.push(letter) if letter && letter.vegeterian
+        @participants_vegan_application_letters.push(letter) if letter && letter.vegan
+        @participants_omnivorous_application_letters.push(letter) if letter && !letter.vegan && !letter.vegeterian
```
Blaidd-Drwg: No need for the `letter &&`, you already checked for nil

lib/pdf_generation/participants_pdf.rb
```diff
@@ -0,0 +1,87 @@
...
+
+  private
+    def create_summary
+      text t("events.participants.print_title", title: @event.name), size: 20
+      move_down 20
+      text @event.date_ranges[0].to_s, size: 10
+      move_down 20
+      text t("events.participants.print_summary", number: @event.participants.count), size: 15
+      
+      if @participants_omnivorous_application_letters.count > 0
```
Blaidd-Drwg: Shorter: `if @participants_omnivorous_application_letters.any?`

spec/factories/application_letters.rb
```diff
@@ -43,6 +43,16 @@
...
   factory :application_letter_accepted, parent: :application_letter do
     status :accepted
+  end  
+
+  factory :application_letter_accepted_vege, parent: :application_letter do
+    status :accepted
+    vegeterian true
+  end  
+
+  factory :application_letter_accepted_veg, parent: :application_letter do
```
Blaidd-Drwg: > application_letter_accepted_veg

Two more characters won't kill you :wink:

spec/features/participants_eating_habit_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    visit events_path
+
+    click_link 'Bewerben'
+    expect(page).to have_current_path(new_application_letter_path(:event_id => @event.id))
+  end
+
+
+
+
+
```
Blaidd-Drwg: That's a lot of empty lines :stuck_out_tongue_winking_eye:

Blaidd-Drwg (changes requested):

lib/pdf_generation/participants_pdf.rb
```diff
@@ -0,0 +1,87 @@
...
+        text t("events.participants.print_summary_omnivorous", number: @participants_omnivorous_application_letters.count), size: 15
+        @participants_omnivorous_application_letters.each do |p|
+          text p.user.name, size: 10
+        end
```
Blaidd-Drwg, line 52: According to Coveralls, this block isn't covered by any tests. I know it's trivial but... think of those gorgeous 100%.

app/helpers/participants_overview_helper.rb
Blaidd-Drwg: These two lines aren't covered as well.

app/views/profiles/show.html.erb
radscheit: This wasn't my idea, i think i merged it in.

Blaidd-Drwg: You're right, looks like the automatic merge messed up. This was removed from the dev [last week](https://github.com/hpi-swt2/workshop-portal/commit/48418a55dba13ec3e101caa2c8e28c1193ad9491), but since you had modified the line it was kept. It can be deleted now :slightly_smiling_face:

lib/pdf_generation/participants_pdf.rb
```diff
@@ -0,0 +1,87 @@
...
+      if @participants_vegan_application_letters.count.any?
+        text t("events.participants.print_summary_vegan", number: @participants_vegan_application_letters.count), size: 15
+        @participants_vegan_application_letters.each do |p|
+          text p.user.name, size: 10
+        end
+      end
+
+      move_down 20
+
+      if @participants_vegeterian_application_letters.count.any?
```
Blaidd-Drwg: Without the `count`, just `if @participants_vegeterian_application_letters.any?`
You're asking the collection if it contains any elements :wink:

app/helpers/applicants_overview_helper.rb
```diff
@@ -11,8 +11,20 @@ def sort_caret(label, attr)
...
+    
+    if params[:sort] && params[:sort] != 'applicant_age_when_event_starts' && params[:sort] != 'eating-habits'
+      @application_letters.sort_by! {|l| l.user.profile.send(params[:sort]) } 
+    end
+
+    if params[:sort] && params[:sort] == 'applicant_age_when_event_starts' && params[:sort] != 'eating-habits'
+      @application_letters.sort_by! {|l| l.send(params[:sort]) } 
+    end
+    
+    if params[:sort] && params[:sort] == 'eating-habits'
```
Blaidd-Drwg: I'm confused by the conditionals.
In line 23 and 19, why do you check if `params[:sort]` is not nil?
In line 19, why do you check if `params[:sort]` equals `eating-habits`?
Line 15: if `params[:sort]` is neither `applicant_age_when_event_starts` nor `eating-habits`, what is it then?

Also, using `send` with a parameter set by the user is potentially dangerous because it allows him to call any function.

config/locales/de.events.yml
```diff
@@ -56,6 +56,13 @@ de:
       unnecessary: "nicht benötigt"
       available: "vorhanden"
       unavailable: "nicht vorhanden"
+      print_habits: "Essgewohnheiten drucken (PDF)"
+      print_title: "Essgewohnheiten der Teilnehmer für %{title}"
+      print_summary: "Insgesamt werden wir bei dieser Veranstaltung %{number} Teilnehmer/-innen haben."
+      print_summary_vegetarian: "%{number} Personen ernähren sich vegetarisch."
+      print_summary_vegan: "%{number} Personen ernähren sich vegan."
+      print_summary_omnivorous: "%{number} Personen ernähren sich omnivor."
```
Blaidd-Drwg: Doesn't sound right if `number` is 1. I18n has a feature for adding separate singular and plural translations, you could use that.

Blaidd-Drwg (changes requested):
Almost done, just hang in there :wink:

app/helpers/applicants_overview_helper.rb
radscheit: Its outdated. See my last Commit.

config/locales/de.events.yml
radscheit: Done.

Blaidd-Drwg (approved):

PR #263, opened 2017-01-07
Add support for markdown. Fixes #258.

### LINKED ISSUE #258: Markdown (first on redesigned event details page)
- [ ] Markdown rendering support (gem)
- [ ] should be able to render the markdown provided in creation (#238) (The whole block in the center should be the rendered description)
- [ ] Renderer is invoked for description of event details page

tom95:
Adds redcarpet (markdown renderer) as a dependency.

The description on the event page will now be rendered in markdown. Descriptions in event lists will be truncated by first removing all elements except paragraphs and links and then truncating to make sure we don't kill any formatting by cutting at the wrong spot. If it turns out that people always start their descriptions with lists, we will probably have to be a bit more sophisticated here.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,7 +1,43 @@
+require 'redcarpet'
+require 'redcarpet/render_strip'
+
+# Very barebone "markdown" renderer that ignores all elements
+# except paragraphs
+class MarkdownRenderTruncate < Redcarpet::Render::Base
+  def paragraph(text)
+    text + '<br>'
+  end
+end
+
 module ApplicationHelper
   def menu_items
     (menu_item t(:events, scope: 'navbar'), events_path) +
     (menu_item t(:requests, scope: 'navbar'), requests_path)
+  end
+
+  # Render the given string as markdown
+  #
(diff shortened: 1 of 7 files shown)


app/helpers/application_helper.rb
```diff
@@ -1,9 +1,49 @@
...
     (menu_item t(:events, scope: 'navbar'), events_path) +
     (menu_item t(:requests, scope: 'navbar'), requests_path)
   end
 
+  # Render the given string as markdown
+  #
+  # @param text String containing markdown
+  # @param truncatable boolean leave only paragraph elements
+  # @return string html_safe rendered markdown
+  def markdown(text, truncatable = false)
```
MasterCarl: I would name `truncatable` differently, maybe `should_truncate` or simply `truncate`

tom95: Well it doesn't truncate anything, it just creates a string suitable for truncation :P I intentionally left out the actual truncation because rails' `truncate` function has various options on how the truncation should happen and I didn't want to limit those (e.g. where the "read on" link is supposed to point to).

PR #265, opened 2017-01-07
218 1.22 Anfragen overview page

### LINKED ISSUE #218: US_1.22: Anfragen Overview page
**As** 
organizer/coach
 
**I want to** 
have an overview page where I see all existing requests for workshops with the topic wish, the name of the person that made the request, the date wish and the and amount of people that would participate. I also want to see a status that shows if the request is open or accepted. I want a details button!

**in order to** 
have better overview

** 🍪**
- [ ]

MasterCarl:
Fixes #218


spec/controllers/requests_controller_spec.rb
```diff
@@ -38,7 +38,7 @@
     describe "GET #index" do
       it "assigns all requests as @requests" do
         get :index, session: valid_session
-        expect(assigns(:requests)).to eq([@a_request])
+        expect(response).to redirect_to(root_url)
```
tom95, line 41: Looking at the comment for `valid_session` ("This should return the minimal set of values that should be in the session in order to pass any filters (e.g. authentication)") I would say this test shouldn't have been changed, but the session should have been adapted. If adapting the session seems overly tedious (which seems like it could be the case looking at [this](https://github.com/plataformatec/devise/wiki/How-To:-Test-controllers-with-Rails-3-and-4-(and-RSpec)#rails-3-rspec-scaffolds)), I guess changing the title of the test case could be sufficient as well.

tom95 (changes requested):
Great changes! Thanks for also adapting the navbar as well :)
Only a (hopefully) small fix on the tests missing to make it perfect.

spec/controllers/requests_controller_spec.rb
MasterCarl: I had already tried to make it work with `valid_session`, but couldn't get it to work (and I couldn't find an example of how to do this with `cancancan`, which handles permissions). I do think it would be nice not to change this test, but it kind of is already covered in the feature spec.
I did forget to rename the test, though.

tom95 (approved):

PR #268, opened 2017-01-09
38 3.10 save email drafts

timbasel:
Refactoring of the email system, allow saving of email templates


spec/factories/email_templates.rb
```diff
@@ -0,0 +1,26 @@
...
+    hide_recipients false
+    subject "Accepted EmailTemplate-Subject"
+    content "Accepted EmailTemplate-Content"
+  end
+
+  factory :email_template_accepted, parent: :email_template do
+    status :accepted
+  end
+
+  factory :email_template_rejected, parent: :email_template do
```
aloeser, line 23: I think it would be a good idea to also change subject and content, it might be confusing otherwise if someone has to deal with this factory in tests

app/models/email.rb
```diff
@@ -1,3 +1,11 @@
+# == Schema Information
```
aloeser: Not sure if we want this here, since Emails are not saved into the DB afaik?
Not sure if its bad either

app/views/emails/email.erb
```diff
@@ -0,0 +1,13 @@
+<%- model_class = Email -%>
+<div class="page-header">
+</div>
+<div>
+  <div class="email-form col-md-8">
+    <%= render partial: 'emails/email_form', locals: { email: @email } %>
+  </div>
+  <div class="email-templates col-md-4">
+    <% if params[:status] %>
+      <%= render partial: 'emails/templates', locals: { template: @templates } %>
```
aloeser: Is it intended to render no templates at all when no status is set?

aloeser (changes requested):
reviewed only the app/ part yet, will review tests later

app/views/emails/email.erb
timbasel: yes, because templates are always linked to a status.

aloeser: The other option would be to render all. Did you talk with @tobiduer about this yet?

timbasel: will do

timbasel: @tobiduer wants a third status for templates 'others' that are displayed if no other status is given, therefore will be changed

spec/views/emails/email_form.html.erb_spec.rb
```diff
@@ -19,9 +23,11 @@
 
   it "renders email send button" do
     expect(rendered).to have_button(I18n.t('emails.email_form.send'))
+    expect(rendered).to have_button(I18n.t('emails.email_form.save_template'))
   end
 
   it "fills recipients_fills with set recipients" do
```
aloeser, line 33: can you fix the test name too? not your fault, but its a one-liner

timbasel: Fixed, but it sure is my fault ;)

app/models/email.rb
timbasel: I was not sure either, but wanted to stay consistent with the other models, also the comment does not refer to a database_table, it just lists the attributes that the model has.

app/assets/stylesheets/email.css
```diff
@@ -1,7 +1,30 @@
+/*
```
frederike-ramin: Why name this file into singular? All the other css files are also plural.

timbasel: will be changed.

app/controllers/emails_controller.rb
```diff
@@ -1,11 +1,87 @@
...
+      EmailTemplate.where(status: EmailTemplate.statuses[:default]).to_a
+    end
+  end
+
+  def get_status
+    if params[:status]
+       params[:status]
+    else
+      :default
+    end
```
frederike-ramin: This could be a one-liner if you use ternary operator.

timbasel: made it that way to stay consistent with the other get_ function above (that is to long for a one-liner). Your decision if I should change it.

app/controllers/emails_controller.rb
```diff
@@ -1,11 +1,87 @@
...
-    redirect_to :events, notice: t('.sending_successful')
+    if @email.valid?
+      Mailer.send_generic_email(@email.hide_recipients, @email.recipients, @email.reply_to, @email.subject, @email.content)
+
+      @event = Event.find(params[:event_id])
+      @event.lock_application_status
+
+      redirect_to @event, notice: t('.sending_successful')
+    else
+      @event = Event.find(params[:event_id])
```
frederike-ramin: @event should still be th same as in show, so you don't have to assign it again, or if you want to, at least to it for both alternatives before the if.

app/controllers/emails_controller.rb
```diff
@@ -1,11 +1,87 @@
...
+  def get_status
+    if params[:status]
+       params[:status]
+    else
+      :default
+    end
+  end
+
+  def send?
+    params[:commit] == t('emails.email_form.send')
```
frederike-ramin: Since when are params localized?! O.O

app/models/ability.rb
```diff
@@ -56,6 +56,7 @@ def initialize(user)
       cannot :update, ApplicationLetter
       # Organizers can view, edit and print Applications, view participants for, upload materials for, print agreement letters for and manage Events
       can [:view_applicants, :edit_applicants, :view_participants, :print_applications, :manage, :upload_material, :print_agreement_letters], Event
+      can :send_email, Email
```
frederike-ramin, line 62: Please add ability test.

config/locales/de.emails.yml
```diff
@@ -2,8 +2,8 @@
 
 de:
   emails:
-    send_email:
-      sending_successful: 'Email wurde erfolgreich versendet'
+    templates:
+      templates: 'Templates'
```
frederike-ramin: How about "Vorlagen"?

config/locales/de.emails.yml
```diff
@@ -13,4 +13,10 @@ de:
       recipients: 'Empfänger'
       subject: 'Betreff'
       content: 'Inhalt'
-      send: 'Senden'
\ No newline at end of file
+      send: 'Senden'
+      save_template: 'Als Template speichern'
```
frederike-ramin: Change accordingly.

spec/controllers/email_controller_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+      end
+    end
+
+    context "with valid rejected applications" do
+      before :each do
+        @application = FactoryGirl.create(:application_letter_rejected, event: @event, user: FactoryGirl.build(:user))
+        @template = FactoryGirl.create(:email_template_rejected)
+      end
+
+      it "sets @email with the email of the accepted application" do
```
frederike-ramin: rejected, not accepted.

spec/controllers/email_controller_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+        get :show, event_id: @event.id, status: :rejected
+        expect(assigns(:email)).to be_a(Email)
+        expect(assigns(:email).recipients).to eq(@application.user.email)
+      end
+
+      it "sets @template with template for accepted emails" do
```
frederike-ramin: rejected, not accepted.

spec/controllers/email_controller_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+        get :show, event_id: @event.id, status: :rejected
+        expect(assigns(:templates)).to eq([@template])
+      end
+    end
+  end
+
+  describe "POST #submit" do
+    context "with valid email pressing the send button" do
+      before :each do
+        @email = FactoryGirl.build(:email).attributes
```
frederike-ramin, line 56: What is the .attributes for?

spec/factories/email_templates.rb
frederike-ramin: I don't agree to be honest. I think that should be fine.

spec/models/email_template_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+#  hide_recipients  :boolean          not null
+#  subject          :string           not null
+#  content          :string           not null
+#
+
+require 'rails_helper'
+
+describe EmailTemplate do
+
+  it "can be created by user factory" do
```
frederike-ramin: user factory?

spec/models/email_template_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+    expect(FactoryGirl.build(:email_template)).to be_valid
+  end
+
+  it "can not be created without manditory fields" do
+    expect(FactoryGirl.build(:email_template, subject: '', content:'')).to_not be_valid
+  end
+
+  it "accepts valid status values" do
+    expect(FactoryGirl.build(:email_template, status: :accepted)).to be_valid
+    expect(FactoryGirl.build(:email_template, status: :rejected)).to be_valid
```
frederike-ramin, line 26: perhaps also test with non-valid status value?

spec/models/event_spec.rb
```diff
@@ -168,16 +168,33 @@
...
+      @application = FactoryGirl.create(:application_letter_accepted, event: @event, user: FactoryGirl.create(:user))
+      expect(@event.email_addresses_of_type(:accepted)).to eq(@application.user.email)
+    end
+
+    it "returns email address of rejected application" do
+      @application = FactoryGirl.create(:application_letter_rejected, event: @event, user: FactoryGirl.create(:user))
+      expect(@event.email_addresses_of_type(:rejected)).to eq(@application.user.email)
+    end
+
+    it "returns email address only of the given type" do
```
frederike-ramin, line 182: With this test, you can just delete the two right above that.

spec/views/emails/email_form.html.erb_spec.rb
```diff
@@ -17,11 +21,17 @@
...
     expect(rendered).to have_button(I18n.t('emails.email_form.send'))
+    expect(rendered).to have_button(I18n.t('emails.email_form.save_template'))
+  end
+
+  it "renders copy recipients button" do
+    expect(rendered).to have_button(I18n.t('emails.email_form.copy'))
   end
 
   it "fills recipients_fills with set recipients" do
+    expect(rendered).to have_field('email_reply_to', with: @email.reply_to)
```
frederike-ramin, line 34: Where is this field? I don't see it when viewing the email show site.

spec/views/emails/template.html.erb_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+
+RSpec.describe "emails/template", type: :view do
+  before(:each) do
+    sign_in(FactoryGirl.create(:user, role: :admin))
+
+    @template = FactoryGirl.create(:email_template)
+    @templates = Array(@template)
+    render :partial => "emails/templates"
+  end
+
```
frederike-ramin, line 11: How about the "Template" Heading?

frederike-ramin (changes requested):
Oh wow..that is a large PR..

app/controllers/emails_controller.rb
aloeser: I think its okay to set @event again, its a private method and might be called from other methods than show later on.
But I agree to move it before the if

app/controllers/emails_controller.rb
aloeser: I think "params[:commit]" is the value(=text) of the submit button in the form. And those should be localized

app/models/email.rb
aloeser: Well for me "Schema" refers to the database schema, but its okay for me when we dont change it

spec/controllers/email_controller_spec.rb
aloeser: Why rejected here?

spec/factories/email_templates.rb
aloeser: Nevermind, when i commented the default values in :email_template were like "Subject accepted" and "content accepted", so it either changed or i looked at the wrong diff (happened already twice today..). Those values are fine

app/assets/javascripts/emails.js
```diff
@@ -3,11 +3,14 @@
 
 jQuery(function() {
     $(document).on('click', "#send-emails-clipboard", function() {
-        $("#email_recipients").select();
-        try {
-           document.execCommand('copy');
-        } catch (err) {
-            console.log('Unable to copy emails to the clipboard');
+        var recipients = document.getElementById('email_recipients');
```
Lando-L: You could use the JQuery selector `$('#email_recipients)` instead

app/controllers/emails_controller.rb
Lando-L: I think there is no need to make this consistent with the other get_ functions and would also suggest the shorter ternary operator

frederike-ramin: I would do so.

spec/controllers/email_controller_spec.rb
frederike-ramin: Sorry...placed it wrong. Thanks.

Lando-L (changes requested):

spec/controllers/email_controller_spec.rb
frederike-ramin: rejected, not accepted.

frederike-ramin: rejected, not accepted.

app/controllers/emails_controller.rb
aloeser: You could write get_templates with 1 line too (I think), or 2 if you want to put the status in a temp var

spec/controllers/email_controller_spec.rb
frederike-ramin: rejected, not accepted.

spec/factories/email_templates.rb
timbasel: Yeah, I changed it quite quickly, so do not worry

spec/controllers/email_controller_spec.rb
frederike-ramin: rejected, not accepted.

app/controllers/emails_controller.rb
timbasel: Unfortunately it is the only way to distinguish between two different buttons inside the same form that I could find.

app/assets/javascripts/emails.js
```diff
@@ -3,11 +3,14 @@
...
+        if(recipients.value.match(/\S/)) {
```
querenker: Why should there be whitespace?

spec/controllers/email_controller_spec.rb
timbasel: It gives only the attributes of the email object as dictionary.
It is needed because the post request does not take the email object

spec/models/email_template_spec.rb
timbasel: copy and paste is bad

app/controllers/emails_controller.rb
querenker: Not tested, but maybe this can work as a one liner. But I think a second line for the status would improve the readability.
`EmailTemplate.where(status: EmailTemplate.statuses[params[:status] ? params[:status].to_sym :  :default ]).to_a`

app/controllers/emails_controller.rb
```diff
@@ -1,11 +1,87 @@
...
+
+    @templates = get_templates
+    @addresses = @event.email_addresses_of_type(get_status)
+
+    @email = Email.new(hide_recipients: true, reply_to: 'workshop.portal@hpi.de', recipients: @addresses,
+                       subject: '', content: '')
+    render :email
+  end
+
+  def submit
```
querenker, line 15: For me this two thinks looks very different and should not be handled by one method?

spec/views/emails/email_form.html.erb_spec.rb
timbasel: The text field 'Absender' should be there, it is visible on my end

app/controllers/emails_controller.rb
```diff
@@ -1,11 +1,87 @@
...
   end
 
-  private
+  def save_template
+    @email = Email.new(email_params)
+
+    @template = EmailTemplate.new({ status: get_status, hide_recipients: @email.hide_recipients,
+                                    subject: @email.subject, content: @email.content })
+
+    if @email.validate_attributes [:subject, :content] and @template.save
```
querenker: Please notice that `and` and `&&` are not synonyms. The `&&` is a boolean operator and the `and` is a control flow operator. Even if both works in this example the `and` can be very confusing in other places.

config/locales/de.emails.yml
timbasel: PO approved that change

querenker (changes requested):
Part 1 of the review, will check the rest later.

app/controllers/emails_controller.rb
timbasel: the problem with this is that you cannot have a second button for the same form that directs to a different function. You can only use "commit" parameter to distinguish the different buttons

app/models/ability.rb
timbasel: done

spec/controllers/email_controller_spec.rb
timbasel: I think here it should be accepted?

spec/views/emails/template.html.erb_spec.rb
timbasel: test added

spec/controllers/email_controller_spec.rb
aloeser: Yes, @frederike-ramin was refering to line 46 (which is already fixed). Either Github is bugged, or those annotations are placed based on the content, it was not possible to attach the comment to line 46, after submitting the comment it always referenced this line

spec/views/emails/template.html.erb_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  it ("renders template headline") do
```
aloeser: I think we never used () around the "" in tests (or at least I havent written or seen this before)

aloeser (changes requested):

spec/controllers/email_controller_spec.rb
frederike-ramin: Yes, @aloeser, thanks for explaining!

spec/models/email_template_spec.rb
frederike-ramin: I meant in a different test. This doesn't really fit in here.

spec/models/event_spec.rb
frederike-ramin: @TimBasel

spec/views/emails/email_form.html.erb_spec.rb
frederike-ramin: I thought that would be a different one? "Absender" is to me different to "Antworten an"..

app/models/email.rb
timbasel: Would the wording "Model Information" be better?

spec/views/emails/email_form.html.erb_spec.rb
timbasel: Yes, but for the end-user it is kind of the same. Because he/she does not know of the internal email address and when you answer to an email you generally reply to the "Absender".

frederike-ramin: Ah well..acceptable I guess..although it's pretty irritating for me.

frederike-ramin: Ah well..acceptable I guess..although it's pretty irritating for me.

spec/models/email_template_spec.rb
timbasel: moved

spec/views/emails/email_form.html.erb_spec.rb
timbasel: should I ask the PO about that?

app/models/email.rb
frederike-ramin: Suppose so.

spec/views/emails/email_form.html.erb_spec.rb
frederike-ramin: Yes maybe, but that's not actually part of this PR, right?

frederike-ramin (approved):

spec/views/emails/email_form.html.erb_spec.rb
timbasel: yes

Lando-L (approved):

app/controllers/emails_controller.rb
querenker: I think [this StackOverflow question](http://stackoverflow.com/questions/3332449/rails-multi-submit-buttons-in-one-form) could help :)

app/assets/javascripts/emails.js
```diff
@@ -3,11 +3,14 @@
 
 jQuery(function() {
     $(document).on('click', "#send-emails-clipboard", function() {
-        $("#email_recipients").select();
-        try {
-           document.execCommand('copy');
-        } catch (err) {
-            console.log('Unable to copy emails to the clipboard');
+        var recipients = $('#email_recipients')
+        if(recipients.val.length == 0) {
```
querenker: I think it would make more sense if the value is greater than 0? Currently it doesn't work on my device....

app/controllers/emails_controller.rb
```diff
@@ -1,11 +1,78 @@
...
+  private
+
   def send_email
     @email = Email.new(email_params)
-    Mailer.send_generic_email(@email.hide_recipients, @email.recipients, @email.reply_to, @email.subject, @email.content)
-    redirect_to :events, notice: t('.sending_successful')
+    @event = Event.find(params[:event_id])
+
+    if @email.valid?
+      Mailer.send_generic_email(@email.hide_recipients, @email.recipients, @email.reply_to, @email.subject, @email.content)
```
querenker: Is it possible to add a send method for the email object with this line?

app/controllers/emails_controller.rb
```diff
@@ -1,11 +1,78 @@
...
+      render :email
+    end
   end
 
-  private
+  def save_template
+    @email = Email.new(email_params)
+
+    @template = EmailTemplate.new({ status: get_status, hide_recipients: @email.hide_recipients,
+                                    subject: @email.subject, content: @email.content })
```
querenker, line 49: Maybe this line could also be in an email method to get the current email as template?

app/views/events/_applicants_overview.html.erb
```diff
@@ -63,20 +63,26 @@
...
       <div class="btn-group pull-right" role="group">
         <%= link_to t('events.badges.print_button_label'), event_badges_path(@event), :class => 'btn btn-default'%>
-        <%= link_to t('.sending_acceptances'), event_send_acceptance_emails_path, class: 'send-emails-button btn btn-default' %>
-        <%= link_to t('.sending_rejections'), event_send_rejection_emails_path, class: 'send-emails-button btn btn-default' %>
+        <% if can? :send_email, Email %>
+          <%= link_to t('.sending_acceptances'), event_email_show_path(@event, status: :accepted), class: 'send-emails-button btn btn-default' %>
+          <%= link_to t('.sending_rejections'),  event_email_show_path(@event, status: :rejected), class: 'send-emails-button btn btn-default' %>
+        <% end %>
       </div>
     <% end %>
```
querenker, line 90: There is some code duplication in this code. Maybe it could be simplified with local variables?

app/controllers/emails_controller.rb
```diff
@@ -1,11 +1,78 @@
...
+  def get_templates
+      EmailTemplate.where(status: EmailTemplate.statuses[get_status]).to_a
+  end
+
+  def get_status
+    params[:status] ? params[:status].to_sym : :default
+  end
+
+  def send?
+    params[:commit] == t('emails.email_form.send')
```
querenker: I already noticed that in a comment, but I think it is not as easy to find. There is [a StackOverflow question](http://stackoverflow.com/questions/3332449/rails-multi-submit-buttons-in-one-form) that could help :)

querenker (changes requested):
Otherwise it looks good for me

app/controllers/emails_controller.rb
```diff
@@ -1,11 +1,78 @@
...
+    @templates = get_templates
+    @addresses = @event.email_addresses_of_type(get_status)
+
+    @email = Email.new(hide_recipients: true, reply_to: 'workshop.portal@hpi.de', recipients: @addresses,
+                       subject: '', content: '')
+    render :email
+  end
+
+  def submit
+    authorize! :send_email, Email
```
aloeser, line 16: is the right to send emails equal to the right to save templates?

app/controllers/emails_controller.rb
```diff
@@ -1,11 +1,78 @@
...
+
   def send_email
     @email = Email.new(email_params)
-    Mailer.send_generic_email(@email.hide_recipients, @email.recipients, @email.reply_to, @email.subject, @email.content)
-    redirect_to :events, notice: t('.sending_successful')
+    @event = Event.find(params[:event_id])
+
+    if @email.valid?
+      Mailer.send_generic_email(@email.hide_recipients, @email.recipients, @email.reply_to, @email.subject, @email.content)
+      @event.lock_application_status
```
aloeser, line 34: right now, this would lock the application status with every email send? (and not only acceptance/rejection mails?)

app/models/email.rb
```diff
@@ -10,14 +18,21 @@ class Email
...
+  validates_presence_of :recipients, :reply_to, :subject, :content
+  validates_inclusion_of :hide_recipients, in: [true, false]
 
   def initialize(attributes = {})
     attributes.each do |name, value|
       send("#{name}=", value)
     end
   end
 
+  def validate_attributes(attributes = [])
```
aloeser: I couldnt find a test for this.
Not really sure what it does too. Why `validates_presence_of`?
Few lines above you also use `validates_inclusion_of`

aloeser (changes requested):

app/controllers/emails_controller.rb
```diff
@@ -1,11 +1,78 @@
...
+    else
+      flash.now[:alert] = t('.saving_failed')
+    end
+    @event = Event.find(params[:event_id])
+    @templates = get_templates
+
+    render :email
+  end
+
+  def get_templates
```
aloeser: Maybe you could turn this into a scope (and move it to the model)?

app/controllers/emails_controller.rb
```diff
@@ -1,11 +1,78 @@
...
+      EmailTemplate.where(status: EmailTemplate.statuses[get_status]).to_a
+  end
+
+  def get_status
+    params[:status] ? params[:status].to_sym : :default
+  end
```
aloeser, line 64: please add a test

app/controllers/emails_controller.rb
aloeser: please add a test

app/controllers/emails_controller.rb
```diff
@@ -1,11 +1,78 @@
...
+
+  def send?
+    params[:commit] == t('emails.email_form.send')
+  end
+
+  def save_template?
```
aloeser: please add a test

aloeser (changes requested):

app/controllers/emails_controller.rb
timbasel: yes, because saving templates is part of the Sending Email Process

app/views/events/_applicants_overview.html.erb
timbasel: It could be done, but I think it would be more complicated. Because you only have two buttons it would just add another line

app/controllers/emails_controller.rb
timbasel: yes, but right now there are only these types of emails

frederike-ramin: Either you do it or we have to change that in our issue.

aloeser (approved):

app/controllers/emails_controller.rb
querenker: @TimBasel What do you think?

app/views/events/_applicants_overview.html.erb
querenker: It is not nice, but I think it is OK

app/controllers/emails_controller.rb
timbasel: The Problem is that because of the status the data comes from 2 different sources, so it would not be simpler

querenker (approved):

Baschdl:
🎉 🎉 🎉

PR #298, opened 2017-01-11
master -> production in README

cmfcmf:


bjrne (approved):
I'm not sure wether Errbit works like that, but it should.

Issue #49, opened 2016-11-18
US_4.15: Unpublish  button & show status 💰 5


Helene0:
**As**
organizer

**I want to**
have an unpublish button and status label for events

**In order to**
manage them better

**Acceptance Criteria:**
[ ] unpublish/publish switch on "event edit"-page & "create event"-page which doesn't delete the event but removes the event from the public/unpublic list of events
[ ] show status of event (publish, unpublish) for organizer

![rastergrafik](https://cloud.githubusercontent.com/assets/8758907/21842628/f6cfa9a6-d7e7-11e6-8802-ffe31c4340be.png)

Label:
<img width="76" alt="label" src="https://cloud.githubusercontent.com/assets/8758907/21842633/fa73be76-d7e7-11e6-92f4-cb274040e972.PNG">

**Tasks**
- [x] add published field to database
- [x] add button to form
- [x] organizers can see event status
- [x] pupils can only see published events
- [x] unify with old draft functionality

Issue #304, opened 2017-01-11
Archive Page For Past Events


hendraet:
**As**
pupil

**I want**
an archive page for past events

**In order to**
be able to browse past events

**Dependencies:**

**Estimate:**

**Acceptance Criteria:**
- [ ] archive page should have the same design as the event overview page (except apply buttons)
- [ ] should be reachable via button on the overview page ("Vergangene Veranstaltungen")
- [ ] button on the left over first event (in the same "line" as the new button for organisers)
- [x] overview page should now be sorted like main page so that the next events are displayed at the top (already done in #316)

tobiasduerschmid:
@tom95: Could you please add sample data for previous events to allow us to test this feature and demonstrate it every time?

bjrne:
@tobiduer It's not that easy and would require a medium sized refactoring

tobiasduerschmid:
@bjrne: Then please propose another way, how we can see that the implementation fits the acceptance criteria.

tobiasduerschmid:
As discussed on Slack, the check that denies creating past events will be removed and the sampled data will be added.

tobiasduerschmid:
@tom95: Could you please add a title / headline ("Veranstaltungs-Archiv") for the archive page?

PR #316, opened 2017-01-11
49 unpublish events

Fabian-Sommer:
(#49)


app/models/event.rb
```diff
@@ -222,7 +222,7 @@ def material_path
     end
   end
 
-  scope :draft_is, ->(draft) { where("draft = ?", draft) }
+  scope :draft_is, ->(published) { where("published = ?", published) }
```
tom95: You should rename this or negate the published

app/models/event.rb
```diff
@@ -231,7 +231,7 @@ def material_path
   # @param only_public Set to true to not include drafts
   # @return List of events
   def self.sorted_by_start_date(only_public)
-    (only_public ? Event.draft_is(false) : Event.all)
+    (only_public ? Event.draft_is(true) : Event.all)
```
tom95: Change from above should be reflected here, so that it makes sense again

app/views/events/_event.html.erb
```diff
@@ -1,3 +1,4 @@
+<% if event.published or can? :view_unpublished, Event %>
```
tom95: I would say this should be filtered out in the controller, starting from the query, and not in the view.

app/views/events/_event.html.erb
```diff
@@ -11,7 +12,8 @@
   <h3>
     <%= event.name %>
     <small>
-      <%= tag_label t('activerecord.attributes.event.draft'), :default if event.draft %>
+      <%= tag_label t('events.notices.is_published'), :success if event.published and can? :view_unpublished, Event %>
+      <%= tag_label t('activerecord.attributes.event.draft'), :default if not event.published and can? :view_unpublished, Event %>
```
tom95: I think the requirement to show unpublished/published on the view came from the old table layout. In the new layout, just showing a draft label should be enough, since any other event is of course published. @hpi-swt2/product-owner-2016

db/migrate/20161126190339_add_draft_to_events.rb
```diff
@@ -1,5 +0,0 @@
-class AddDraftToEvents < ActiveRecord::Migration
-  def change
-    add_column :events, :draft, :boolean
-  end
-end
```
tom95, line 5: You won't want to remove any migrations ever. It also seems you forgot to add your own migration that would add the published field.

tom95 (changes requested):
I don't see why you had to rename the `draft` field to `published`. This change really cluttered the merge request and introduced a couple of mistakes in this MR. Was there a need to rename the field?

db/migrate/20161126190339_add_draft_to_events.rb
bjrne: [This page](http://edgeguides.rubyonrails.org/active_record_migrations.html#running-migrations) has everything you need to know about migrations. I really recommend to read it through.

app/views/events/_event.html.erb
Fabian-Sommer: This request was added to US 3 days ago (at least to title of US)

db/migrate/20170111135212_rename_draft_to_published.rb
```diff
@@ -0,0 +1,5 @@
+class RenameDraftToPublished < ActiveRecord::Migration
+  def change
+    rename_column :events, :draft, :published
```
aloeser, line 12: Dont you need to invert the values too? Otherwise the unpublished (drafts) will become published, and the other way round

aloeser (changes requested):

db/migrate/20170111135212_rename_draft_to_published.rb
Fabian-Sommer: Is that a problem (I repopulate sample data when pulling from dev)?

bjrne: It isn't now, but if our System would be deployed already, it would be.

app/controllers/events_controller.rb
```diff
@@ -7,7 +7,7 @@ class EventsController < ApplicationController
 
   # GET /events
   def index
-    @events = Event.sorted_by_start_date(!can?(:edit, Event)).reverse
+    @events = Event.sorted_by_start_date(!can?(:view_unpublished, Event))
```
bjrne, line 10: Is there a reason you removed the `.reverse`?

db/migrate/20170111135212_rename_draft_to_published.rb
aloeser: Repopulate sample data sounds for me like a database reset. I dont think you want to reset the database with every single commit/merge when running in production mode.

app/controllers/events_controller.rb
aloeser: btw, you could use `cannot?` instead of `!can?`

Fabian-Sommer: I dont think it is good to have the latest (and therefore probably most uninteresting) events shown first and events that are about to begin (and therefore interesting) towards the bottom. Also, this achieves consistency with the start page. 
Is there a reason for showing them in different orders on start page and event page @tom95 @hpi-swt2/product-owner-2016 ?

bjrne: I was just wondering, because this is part of our user story #304 and I was afraid we would loose track if things are changed on the fly in PRs that have little to do with it.

Fabian-Sommer: Since no one is working on #304 yet and it is not really related to an archive page anyway, I suggest just merging it here and deleting it from #304.

app/views/events/_form.html.erb
```diff
@@ -88,16 +88,27 @@
...
   <div class="form-group">
     <div class="col-lg-offset-2 col-lg-10">
-      <% if @event.draft == false %>
-        <%= f.submit t('.update'), :name => 'publish', :class => 'btn btn-primary' %>
+      <% if @event.published %>
+        <%= f.submit t('.update'), :name => 'update', :class => 'btn btn-primary' %>
       <% else %>
-        <%= f.submit t('.publish'), :name => 'publish', :class => 'btn btn-primary' %>
-        <%= f.submit t('.draft'), :name => 'draft', :class => 'btn btn-default' %>
+        <%= f.submit t('.create'), :name => 'create', :class => 'btn btn-primary' %>
```
tom95, line 109: The label "create" currently appear whenever you have an unpublished event, which could be confusing. I think the `if` above should just ask for `@event.new_record?` to determine whether to show create or update.

app/assets/stylesheets/participants.css
```diff
@@ -0,0 +1,3 @@
+#participants_download_submit_button  {
+    float: right;
+}
```
tom95, line 3: I would guess these changes sneaked in through a different branch?

app/controllers/events_controller.rb
tom95: Yes I think we'll tackle ordering as part of #304 once this is merged. Original intent was, since the main page starts in the present, to show the next event (so ascending order) and on the list page use descending order, since we don't want to start looking at all the past events.

tom95 (changes requested):
Thanks for adapting everything! After those two small fixes I'd say we can finally merge this branch.

app/assets/stylesheets/participants.css
Fabian-Sommer: This should have been part of PR #259, but I forgot to push it

tom95 (approved):

aloeser (approved):

db/migrate/20170111135212_rename_draft_to_published.rb
```diff
@@ -0,0 +1,14 @@
+class RenameDraftToPublished < ActiveRecord::Migration
+  def change
+    reversible do |dir|
+      dir.up do
+        Event.all.each {|e| e.toggle!(:draft)}
+      end
+      dir.down do
+        Event.all.each {|e| e.toggle!(:published)}
```
aloeser: I think it should be :draft here too. When i run the migration it works, but when I try a rollback it fails with the error that Event has no published attribute. I guess the rename happens first

PR #323, opened 2017-01-13
189 variable application fields input

tom95:
This adds the input for custom application fields on the event page. Field names are saved in an ordered array in the database so they can easily be output again and mapped to the user entered values, simply by maintaining the same order.

E.g.:
`["Lieblingsfarbe", "Lieblingstier"]` in `Event.custom_application_fields`

would correspond to

`["Blau", "Capybara"]` in `ApplicationLetter.custom_fields`


db/schema.rb
```diff
@@ -110,7 +111,6 @@
     t.integer  "number_of_participants"
     t.string   "knowledge_level"
     t.text     "annotations"
-    t.integer  "status",                 default: 0
```
bjrne: Is this line intended? (Somebody forgot to commit it earlier?)

bjrne, line 114: Is this line intended? (Somebody forgot to commit it earlier?)

tom95: Yeah apparently, reran migrations after dropping and it stayed.

cmfcmf:
> Field names are saved in an ordered array in the database so they can easily be output again and mapped to the user entered values, simply by maintaining the same order.

This is how I solved it for my personal event management system years ago. It caused many errors and problems and was hard to maintain. I know it'd be more work but a proper entity schema would be a lot better. Things that can potentially break your implementation:
- Custom fields are deleted after somebody already applied
- Sort order of custom fields is changed after somebody applied
- Another custom field is added after somebody applied.

Just my two cents.

cmfcmf:
For the record, my system has the following schema:
1 event has n questions.
1 event has n applications
1 application has n answers
1 answer has 1 question, 1 question has n answers

cmfcmf:
You also run into problems once you extend the custom fields to have parameters themselves: Whether or not a field is editable, required, only allows integers, etc. Of course you can then store arrays of arrays of field options, but it only gets more and more messy.

tom95:
In order to avoid all of those problems our PO decided that all fields are immutable after creation. Using a more sophisticated data structure later on will only require changing very little code from this MR, so I think we can go with this smaller solution for the time being.

cmfcmf:
> In order to avoid all of those problems our PO decided that all fields are immutable after creation. Using a more sophisticated data structure later on will only require changing very little code from this MR, so I think we can go with this smaller solution for the time being.

I see. So when you create an event, you must be really careful not to add any wrong fields or include typos. If that's what's agreed upon for now, go ahead :)

app/models/event.rb
```diff
@@ -17,6 +17,8 @@ class Event < ActiveRecord::Base
   UNREASONABLY_LONG_DATE_SPAN = 300
   TRUNCATE_DESCRIPTION_TEXT_LENGTH = 250
 
+  serialize :custom_application_fields, Array
```
MasterCarl, line 20: Very elegant solution

app/views/events/_show_event.html.erb
```diff
@@ -20,6 +20,16 @@
   <dd><%= @event.knowledge_level %></dd>
   <dt><strong><%= model_class.human_attribute_name(:application_deadline) %>:</strong></dt>
   <dd> <%= I18n.l(@event.application_deadline) + ' ' + I18n.t('midnight') %></dd>
+  <% if can? :manage, Event %>
```
MasterCarl, line 27: `:manage` is the broadest permission, maybe `:create` would be more appropriate. Currently, it wouldn't make much of a difference, though.

spec/features/event_spec.rb
```diff
@@ -176,6 +176,38 @@
...
+      click_button I18n.t(".events.form.publish")
+
+      expect(page).to have_text("Lieblingsfarbe")
+      expect(page).to have_text("Lieblings 'Friends' Charakter")
+    end
+
+    it "should not allow adding fields after event creation" do
+      event = FactoryGirl.create(:event)
+      visit edit_event_path(event)
+      expect(page).to_not have_text("Neues Feld hinzufügen")
```
MasterCarl: This should be independent of translations

MasterCarl (changes requested):
Good work in general! I'd request just that one change.

MasterCarl (approved):

PR #326, opened 2017-01-13
233 3.3 event states

frederike-ramin:
Make event states more obvious by implementing a function that returns them.

--> needs to be used instead of the implicit request


app/models/event.rb
```diff
@@ -163,6 +171,17 @@ def lock_application_status
     update(application_status_locked: true)
   end
 
+  # Returns the current statse of the event (draft-, application-, selection- and execution-phase)
```
Lando-L: Simple typo in state (statse)

spec/models/event_spec.rb
```diff
@@ -188,4 +188,45 @@
...
+
+  it "is in draft phase" do
+    event = FactoryGirl.build(:event)
+    event.draft = true
+    expect(event.phase).to eq(:draft)
+  end
+
+  it "is in application phase" do
+    event = FactoryGirl.build(:event)
+    event.draft = false
```
Lando-L: Not absolute necessary because `event.draft = false` is default case

spec/models/event_spec.rb
```diff
@@ -188,4 +188,45 @@
...
+    event.application_deadline = Date.tomorrow
+    expect(event.phase).to eq(:application)
+  end
+
+  it "is in selection phase" do
+    event = FactoryGirl.build(:event)
+    event.draft = false
```
Lando-L: Same es line number 200

Lando-L: Or you could create a factory for each state of the event

Lando-L (changes requested):

spec/models/event_spec.rb
aloeser: True, but I dont think its good to rely on the default values in factories, since they might be changed

app/models/application_letter.rb
```diff
@@ -28,15 +28,15 @@ class ApplicationLetter < ActiveRecord::Base
 
   enum status: {accepted: 1, rejected: 0, pending: 2, alternative: 3}
   validates :status, inclusion: { in: statuses.keys }
-  
+
```
bjrne, line 32: unnecessary whitespace change

bjrne (approved):
👍

timbasel (approved):

Lando-L (approved):

Issue #278, opened 2017-01-10
US_5.9: Download material 💰 3


julisa99:
**As**
organizer and coach

**I want to**
download the uploaded material

**In order to**
use it for the camp

**Attention**
- #36 was "upload material"

**Acceptance criteria**
- only the coach and organizer can download the material

PR #329, opened 2017-01-14
278 5.9 download material

pinussilvestrus:
- #278  an organizer or a coach can now download files of one event's material area


spec/controllers/events_controller_spec.rb
```diff
@@ -239,6 +239,31 @@
...
+    end
+
+    after :each do
+      filepath = File.join(@event.material_path, @file.original_filename)
+      File.delete(filepath) if File.exist?(filepath)
+    end
+
+    it "download a file from the event's material directory" do
+      post :download_material, event_id: @event.to_param, session: valid_session, file: @file.original_filename
+      response.header['Content-Type'].should eql 'application/pdf'
```
Fabian-Sommer: This causes a deprecation warning whenever running tests. Use expect(...).to eq (...) instead.

Blaidd-Drwg:
Cans for controller missing

app/controllers/events_controller.rb
```diff
@@ -188,6 +188,17 @@ def upload_material
...
+  # POST /events/1/download_material
+  def download_material
+    event = Event.find(params[:event_id])
+    file_full_path = File.join(event.material_path, params[:file])
+    unless File.exists?(file_full_path)
+      redirect_to event_path(event), alert: t("events.material_area.download_file_not_found")
+      return false
+    end
+    send_file file_full_path, :x_sendfile => true
+  end
```
Fabian-Sommer, line 211: You should check if the user has permission to download the file before sending it.

Fabian-Sommer (changes requested):

Fabian-Sommer (approved):

Issue #119, opened 2016-12-02
US_5.6: Group accepted users by color 💰 5


julisa99:
**As**
organizer

**I want to**
group the participants of one event

**in order to**
split the participants in little work groups.

## Acceptance criteria:
- Opportunity for 10 colors (light blue, dark blue, red, yellow, orange, light green, dark green, violett, pink, grey) in a dropdown OR every color with a color picker
- Opportunity to change the color during the camp 
- As organizer I want to label a specific user with an color on the participants page
- As organizer I want to see the column "group" and there I can see rectangle colors for each participant
- As organizer I want to sort by group colors to generate name badges for example. (#33)

##
- [x] Add group to model
- [x] Show on participants page
- [x] Implement color selector
- [x] Make selector save selection to db
- [x] Add support for sorting by group and agreement letter state

h0rnyc0rny:
How much colors?

PR #340, opened 2017-01-17
119 5.6 group accepted users by color

phnm:
US_5.6 (connects to #119)
Implemented the ability to set and sort by groups with colors on the participants page.


app/assets/javascripts/participant_group.js
```diff
@@ -0,0 +1,2 @@
+// Place all the behaviors and hooks related to the matching controller here.
+// All this logic will automatically be available in application.js.
```
Blaidd-Drwg: This file is superfluous

app/assets/stylesheets/participant_group.css
```diff
@@ -0,0 +1,4 @@
+/*
+  Place all the styles related to the matching controller here.
+  They will automatically be included in application.css.
+*/
```
Blaidd-Drwg: Superfluous file

app/controllers/participant_groups_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+    if @participant_group.update_attributes(group_param)
+      redirect_to :back, notice: I18n.t('participant_groups.update.successful') rescue ActionController::RedirectBackError redirect_to root_path
+    else
+      redirect_to :back, alert: I18n.t('participant_groups.update.failed') rescue ActionController::RedirectBackError redirect_to root_path
+    end
+  end
+
+  private
+
+    def group_param
```
Blaidd-Drwg: I would still use the plural `group_params`, since it's a convention.

app/controllers/participant_groups_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  load_and_authorize_resource
+
+  # PATCH/PUT /participant_group/1/group
+  def update
+    if @participant_group.update_attributes(group_param)
+      redirect_to :back, notice: I18n.t('participant_groups.update.successful') rescue ActionController::RedirectBackError redirect_to root_path
+    else
+      redirect_to :back, alert: I18n.t('participant_groups.update.failed') rescue ActionController::RedirectBackError redirect_to root_path
+    end
```
Blaidd-Drwg, line 15: Instead of rescuing the same error twice you can do
```ruby
begin
  if @participant_group.update_attributes(group_param)
    redirect_to :back, notice: I18n.t('participant_groups.update.successful')
  else
    redirect_to :back, alert: I18n.t('participant_groups.update.failed') 
  end
rescue ActionController::RedirectBackError
  redirect_to root_path
end
```

app/helpers/participant_group_helper.rb
```diff
@@ -0,0 +1,2 @@
+module ParticipantGroupHelper
+end
```
Blaidd-Drwg: Superfluous file

app/models/application_letter.rb
```diff
@@ -15,6 +15,7 @@ class ApplicationLetter < ActiveRecord::Base
   belongs_to :event
 
   has_many :application_notes
+  has_one :participant_group
```
Blaidd-Drwg: I'm still not super sure about this. I'll ask what the rest of the team thinks.

app/models/event.rb
```diff
@@ -94,6 +94,18 @@ def participants
     accepted_applications.collect { |a| a.user }
   end
 
+  # Returns the participant group for this event for a given participant (user). If it doesn't exist, it is created
+  #
+  # @param user [User] the user whose participant group we want
+  # @return [ParticipantGroup] the user's participant group
+  def participant_group_for(user)
+    application_letter = self.application_letters.where(user: user).take
```
Blaidd-Drwg: Shorter: `self.application_letters.find_by(user: user)`

app/models/event.rb
```diff
@@ -94,6 +94,18 @@ def participants
...
+    if application_letter.participant_group.nil?
+      application_letter.create_participant_group(group: 0)
+    end
+    self.application_letters.where(user: user).take.participant_group
```
Blaidd-Drwg: Instead of `self.application_letters.where(user: user).take` again, which would mean another database query, you could use the variable `application_letter`. To cover the cases where it is `nil`, you would need to assign the newly created ParticipantGroup to the `application_letter` in the if block.

app/models/event.rb
```diff
@@ -94,6 +94,18 @@ def participants
...
   end
 
+  # Returns the participant group for this event for a given participant (user). If it doesn't exist, it is created
+  #
+  # @param user [User] the user whose participant group we want
+  # @return [ParticipantGroup] the user's participant group
+  def participant_group_for(user)
+    application_letter = self.application_letters.where(user: user).take
+    if application_letter.participant_group.nil?
+      application_letter.create_participant_group(group: 0)
```
Blaidd-Drwg: Since you set 0 as `ParticipantGroup::GROUPS.default`, you can use that instead of the magic number.

app/views/events/_colorselect.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= form_for participant_group, url: participant_group_path(participant_group.id), method: :patch do |f| %>
+  <%= f.select( :group, options_for_select(ParticipantGroup::GROUPS.map { |m| [I18n.t("participant_groups.options.#{m.last}"), m.first.to_i, {'data-color' => '#' + m.last}] }, selected: participant_group.group), {}, { onchange: 'this.form.submit()', class: 'colorselector', autocomplete: 'off' }) %>
```
Blaidd-Drwg: This line is really long, please break it up a bit.
In the `map` block you can directly use key and value if you give it two parameters, like `{ |key, value| [I18n.t("participant_groups.options.#{value}"), key, {'data-color' => '#' + value}] }`

app/views/events/participants.html.erb
```diff
@@ -23,7 +24,10 @@
         <% end %>
       </tr>
 
-      <% @participants.each do |participant| %>
+      <%
+      @participants.sort_by! {|l| @event.participant_group_for(l).send(params[:sort]) } if params[:sort]
+      @participants.reverse! if params[:order] == 'descending'
```
Blaidd-Drwg: Any sorting should happen in the controller

app/views/events/participants.html.erb
```diff
@@ -23,7 +24,10 @@
...
+      @participants.select {|l| @event.participant_group_for(l) }.each do | participant | %>
```
Blaidd-Drwg: Since `participant_group_for` never returns `nil`, this is not needed.

app/views/events/participants.html.erb
```diff
@@ -41,24 +45,43 @@
             <span class="text-danger"><%= t '.unavailable' %></span>
           <% end %>
         </td>
+        <td>
+          <%= render 'colorselect', participant_group: @event.participant_group_for(participant) %>
+        </td>
         <% if can? :print_agreement_letters, Event and @has_agreement_letters %>
           <td>
-            <%= check_box_tag('selected_participants[]', participant.id, false, class: 'participant_selection') %>
+            <%= check_box_tag('selected_participants[]', participant.id, false, { class: 'participant_selection' }) %>
```
Blaidd-Drwg: This change isn't necessary, right?

app/views/events/participants.html.erb
```diff
@@ -41,24 +45,43 @@
...
-    <% if can? :print_agreement_letters, Event and @has_agreement_letters %>
-      <div id="participants_download_submit_button">
-      <%= submit_tag t('events.agreement_letters_download.download_all_as'), name: "download_all", class: "btn btn-default" %>
-      <%= select_tag :download_type, options_for_select([['pdf'], ['zip']]) %>
-      </div>
+    <%= form_tag(event_path(@event) + "/participants/agreement_letters", { method: :post, id: 'downloadLettersForm' }) do %>
+
+      <% if can? :print_agreement_letters, Event and @has_agreement_letters %>
+        <div id="participants_download_submit_button">
+        <%= submit_tag t('events.agreement_letters_download.download_all_as'), name: "download_all" %>
```
Blaidd-Drwg: You removed the class here :wink:

db/migrate/20170110015814_create_participant_groups.rb
```diff
@@ -0,0 +1,8 @@
+class CreateParticipantGroups < ActiveRecord::Migration
+  def change
+    create_table :participant_groups do |t|
+      t.references :application_letter, index: true, foreign_key: true
+      t.integer :group
```
Blaidd-Drwg: In the documentation you put these fields as `not null`, but in the database they are currently not.

Blaidd-Drwg (changes requested):
I didn't review the tests yet, but this should already give you something to do in the meantime :wink:

app/controllers/participant_groups_controller.rb
phnm: Are you sure? I thought participant_group_params should be plural, since it would be for all/several parameters of ParticipantGroup, and group_param would be singular, since it's only for the group parameter.

app/views/events/participants.html.erb
phnm: Unless you need several html attributes, which in the end I didn't, so you're right ;)

app/views/events/participants.html.erb
phnm: good spotting! :wink:

spec/controllers/participant_groups_controller_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+  let(:valid_attributes) { FactoryGirl.build(:participant_group).attributes }
+  let(:invalid_attributes) { FactoryGirl.build(:participant_group, event: nil).attributes }
+
+  # This should return the minimal set of values that should be in the session
+  # in order to pass any filters (e.g. authentication) defined in
+  # ApplicationsController. Be sure to keep this updated too.
+  let(:valid_session) { {} }
+
+  before(:each) do
+    request.env['HTTP_REFERER'] = root_url
```
Blaidd-Drwg: `root_url` is also the path the user gets redirected to in case of a RedirectBackError, it would be better using a different path.

spec/factories/participant_groups.rb
```diff
@@ -0,0 +1,7 @@
+FactoryGirl.define do
+  factory :participant_group do
+    event_id 1
+    user_id 1
```
Blaidd-Drwg: This could lead to problems if Events or Users with that id don't exist, or if User 1 isn't actually a participant for Event 1.

spec/features/events_spec.rb
```diff
@@ -188,24 +188,24 @@
...
+  scenario "logged in as Organizer when I want to download agreement letters but no participants are selected, it displays error message", js: true do
     login(:organizer)
     @event = FactoryGirl.create(:event_with_accepted_applications_and_agreement_letters)
     visit event_participants_path(@event)
     click_button I18n.t "events.agreement_letters_download.download_all_as"
     expect(page).to have_text(I18n.t "events.agreement_letters_download.notices.no_participants_selected")
   end
 
-  scenario "logged in as Organizer when I want to download agreement letters but no participants have agreement letters, it displays error message" do
+  scenario "logged in as Organizer when I want to download agreement letters but no participants have agreement letters, it displays error message", js: true do
```
Blaidd-Drwg, line 199: Is the js necessary in these tests?

spec/features/participants_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    @event = FactoryGirl.create(:event)
+  end
+
+  scenario "logged in as Organizer I can sucessfully select a group for a participant", js: true do
+    login(:organizer)
+    @user = FactoryGirl.create(:user)
+    @profile = FactoryGirl.create(:profile, user: @user)
+    @application_letter = FactoryGirl.create(:application_letter_accepted, user: @user, event: @event)
+    @participant_group = FactoryGirl.create(:participant_group, user: @user, event: @event)
+    visit "/events/#{@event.id}/participants"
```
Blaidd-Drwg: Better: `event_participants_path(@event)`

spec/features/participants_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+    select I18n.t("participant_groups.options.#{ParticipantGroup::GROUPS[9]}"), from: "participant_group[group]", match: :first, visible: false
+    expect(page).to have_text(I18n.t("participant_groups.update.successful"))
+  end
```
Blaidd-Drwg, line 17: Can you also test whether the correct group is still selected in the end?

spec/features/participants_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  scenario "logged in as Organizer I can see a table with the participants and sort them by group" do
+    login(:organizer)
+    for i in 1..5
+      user = FactoryGirl.create(:user)
+      profile = FactoryGirl.create(:profile, user: user, last_name: i.to_s)
+      application_letter = FactoryGirl.create(:application_letter_accepted, user: user, event: @event)
+      participant_group = FactoryGirl.create(:participant_group, user: user, event: @event, group: i)
+    end
+
+    visit "/events/#{@event.id}/participants"
```
Blaidd-Drwg: `event_participants_path(@event)`

spec/features/participants_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    table = page.find('table')
+    @event.participants.each do |participant|
+      expect(table).to have_text(participant.profile.name)
+    end
+
+    link_name = I18n.t("activerecord.attributes.participant_group.group")
+    click_link link_name
+    sorted_by_group = @event.participants.sort_by {|p| @event.participant_group_for(p).send("group") }
```
Blaidd-Drwg: Instead of `.send("group")` you can just do `.group` :wink:

spec/views/events/participants.html.erb_spec.rb
```diff
@@ -48,7 +48,31 @@
...
+  it "displays select element with all options" do
+    @user = FactoryGirl.create(:user)
+    @profile = FactoryGirl.create(:profile, user: @user)
+    @event = FactoryGirl.create(:event)
+    @application_letter = FactoryGirl.create(:application_letter_accepted, user: @user, event: @event)
+    @participant_group = FactoryGirl.create(:participant_group, user: @user, event: @event)
+    assign(:event, @event)
+    assign(:participants, @event.participants)
+    render
+    expect(rendered).to have_select('participant_group_group', options: ParticipantGroup::GROUPS.map { |m| I18n.t("participant_groups.options.#{m.last}") })
```
Blaidd-Drwg: Better use |key, value| when mapping over a Hash

Blaidd-Drwg (changes requested):

spec/features/events_spec.rb
phnm: Yes, because the workaround of the nested form conflict is done with js.

app/controllers/participant_groups_controller.rb
Blaidd-Drwg: Well, `participant_group_params` doesn't exist anymore, and `group_param` is for all the parameters, there just happens to be only one :grin: 
So I would rename it to `participant_group_params`, but that's just as an advice, I will still accept the PR :wink:

spec/features/events_spec.rb
Blaidd-Drwg: Alright :slightly_smiling_face:

app/views/events/participants.html.erb
Blaidd-Drwg: @phnm Just move these lines into the controller and you're good to merge :wink:

Blaidd-Drwg: Oh, hold on, also: the `send(params[:sort])` is insecure since it allows the user to execute any function of the ParticipantGroup.

phnm: Oh yeah! This also needs to be fixed in the ApplicantsOverviewHelper module.

Blaidd-Drwg (approved):

Issue #193, opened 2016-12-14
US_4.3a: Email lists for mailing to a few participants 💰 3


julisa99:
**As**
organizer 

**I want to**
generate an email-list with all participants **or some groups** / **single persons** during the camp

**in order to**
avoid coping the email addresses.

**Attention**
Ask Team Dürschmid for mail UI und backend

**Acceptance Criteria**
- The sender of the Email can be a single person (e.g. lisa.ihde@student.hpi.de) or a mailing list (e.g. mintcamp2016@hpi.de)
- The receiver is a group or a single person
- Please ensure, that this will not lead to access violation
- participants overview with button to generate an email-list with all participants **or some groups (only team blue, etc.)**  or a **single person**
- the organizer can select the persons with checkboxes and then click the button to send an e-mail, in the next step, as organizer, I should write the email

tobiasduerschmid:
@Raandom: Where can we find the button for sending the emails after selecting the receivers?

julisa99:
Der Haken für "alle auswählen" ist verpixelt? Lässt sich das verbessern?
<img width="598" alt="haken" src="https://cloud.githubusercontent.com/assets/8758907/22629636/f3489642-ebe9-11e6-8e70-e0055933abf3.PNG">

Bitte setze den Button zum Email versenden rechts neben "Namensschilder ausdrucken".

Blaidd-Drwg:
Die Checkbox wird vom Browser bereitgstellt, und deiner hat anscheinend entschieden, dass Vektorgrafiken nur was für Nerds sind.
Man könnte a) selber eine Vektorgrafik für diese eine Checkbox liefern (und hoffen, dass es in allen Browsern ok aussieht), oder b) die Checkbox kleinermachen.

tobiasduerschmid:
@Raandom @Blaidd-Drwg: Who is the real assignee for this US?

In my Browser (Firefox), the layout seem to be broken. 
![bild1](https://cloud.githubusercontent.com/assets/16103359/22637836/af0eb456-ec44-11e6-8c06-1b82c2a1cbf8.png)
The "alle auswählen" checkbox is huge and should be directly beside the text.

Blaidd-Drwg:
@tobiduer: @Raandom is, but I've spent so much time reviewing it that by now I'm pretty familiar with it :smile:

Raandom:
I'll have a look at that checkbox, I'll probably end up just making it small. Thanx for the review gentlemen and (women).

[...]

PR #342, opened 2017-01-18
193 4.3b real email lists for participants

Raandom:
(#193)


app/assets/javascripts/events.js
```diff
@@ -12,6 +12,26 @@ jQuery(function() {
         modal.find('#send-emails-mailto').attr('href', 'mailto:' + list);
         modal.find('#send-emails-list').val(list);
     });
+
+    var users = $('#users');
```
radscheit: Avoid `$` for `JQuery` in bigger systems. It leads to complications.

app/views/events/participants.html.erb
```diff
@@ -51,12 +52,14 @@
     </table>
 
     <button class="btn btn-default" type="button" data-toggle="modal" data-target="#print_participant_modal" id="open_print_modal">Teilnehmerliste (PDF)</button>
```
radscheit: How about a Bootstrap button group?

app/views/events/participants.html.erb
```diff
@@ -51,12 +52,14 @@
...
+    <button type="button" class="btn btn-default" data-toggle="modal" data-target="#send_participant_email_modal"><%= t('.send_email') %></button>
     <% if can? :print_agreement_letters, Event and @has_agreement_letters %>
       <div id="participants_download_submit_button">
-      <%= submit_tag t('events.agreement_letters_download.download_all_as'), name: "download_all", class: "btn btn-default" %>
+      <%= submit_tag t('events.agreement_letters_download.download_all_as'), name: "download_all", class: "btn btn- default" %>
       <%= select_tag :download_type, options_for_select([['pdf'], ['zip']]) %>
```
radscheit: Why did you add a space in the btn-default class?

config/locales/de.events.yml
```diff
@@ -47,12 +47,21 @@ de:
       unnecessary: "nicht benötigt"
       available: "vorhanden"
       unavailable: "nicht vorhanden"
+      send_email: "Email an Teilnehmer versenden"
```
radscheit: Duden: E-Mail

config/locales/de.events.yml
```diff
@@ -47,12 +47,21 @@ de:
...
         no_participants_selected: "Keine Teilnehmer ausgewählt"
         no_agreement_letters: "Keiner der gewählten Teilnehmer hat eine Einverständniserklärung"
       select_all: "Alle auswählen"
       download_all_as: "Einverständniserklärungen herunterladen als"
+    email_modal:
+      send_email: "E-Mail an Teilnehmer senden"
+      select_users: "Nutzer auswählen"
+      select_groups: "Gruppen auswählen"
+      select_all: "Alle"
+      sender: "Absenderaddresse"
```
radscheit: Absenderadresse

vendor/assets/stylesheets/multi-select.css
```diff
@@ -0,0 +1,93 @@
+.ms-container{
```
radscheit: This should be added to the .less-source.

app/models/event.rb
```diff
@@ -163,13 +163,71 @@ def generate_acceptances_email
...
+  # @return []
+  def groups_with_id
+    []
+  end
+
+  # No functionality because no groups are there yet.
+  # Returns a list of email addresses of all participants that are in one of the given groups
+  #
+  # @return [String]
+  def email_addresses_of_groups(groups)
```
radscheit: groups are already imlemented

radscheit (changes requested):

app/controllers/emails_controller.rb
```diff
@@ -23,6 +24,11 @@ def submit
     end
   end
 
+  def submit_generic
+    @send_generic = true
```
Blaidd-Drwg: Why do you set `@send_generic` here and in `submit`? Where is it used?

app/controllers/emails_controller.rb
```diff
@@ -55,6 +61,17 @@ def send_email
     end
   end
 
+  def send_generic
```
Blaidd-Drwg: I'd rename `send_email` to something more descriptive like `send_acceptance_rejection_email`, in order to make the difference between it and `send_generic` clearer.

Blaidd-Drwg: Same with `submit` and `submit_generic`

app/controllers/emails_controller.rb
```diff
@@ -23,6 +24,11 @@ def submit
     end
   end
 
+  def submit_generic
+    @send_generic = true
+    @templates = []
+    send_generic
+  end
```
Blaidd-Drwg: You don't check if the user is authorized to send mails and you don't distinguish between sending a mail and saving a draft (see `submit`)

app/controllers/emails_controller.rb
```diff
@@ -55,6 +61,17 @@ def send_email
     end
   end
 
+  def send_generic
+    @email = Email.new(email_params)
+    if @email.valid?
+      @email.send_email
+      redirect_to :events, notice: t('emails.submit.sending_successful')
```
Blaidd-Drwg: Could you redirect to the event page the user came from?

app/controllers/events_controller.rb
```diff
@@ -125,6 +125,15 @@ def accept_all_applicants
     redirect_to event_path(event)
   end
 
+  # GET /events/1/send-participants-email
+  def send_participants_email
+    event = Event.find(params[:id])
```
Blaidd-Drwg: Missing authorization

app/models/event.rb
```diff
@@ -177,6 +177,81 @@ def email_addresses_of_type_without_notification_sent(type)
     applications.collect { |a| a.user.email }
   end
 
+  # Returns a string of all email addresses of accepted applications
+  #
+  # @param none
+  # @return [String] Concatenation of all email addresses of accepted applications, seperated by ','
+  def email_addresses_of_accepted_applicants
+    applications = application_letters.where(status: ApplicationLetter.statuses[:accepted])
```
Blaidd-Drwg: `Event` has a `participants` method that does just that

app/models/event.rb
```diff
@@ -177,6 +177,81 @@ def email_addresses_of_type_without_notification_sent(type)
...
+    applications.collect { |a| a.user.email }.join(',')
+  end
+
+  # Returns a new email to a set of participants
+  #
+  # @param criteria String
+  # @param value String
+  # @param sender String
+  # @return [Email] new email
```
Blaidd-Drwg: Missing descriptions of the parameters, formatting.

app/models/event.rb
```diff
@@ -177,6 +177,81 @@ def email_addresses_of_type_without_notification_sent(type)
...
+  def generate_participants_email(all, groups, users, sender)
+    email = Email.new
+    email.hide_recipients = false
+    email.recipients = email_addresses_of_participants(all, groups, users)
+    email.reply_to = sender
+    email.subject = ''
+    email.content = ''
```
Blaidd-Drwg: Is setting the empty strings needed?

app/models/event.rb
```diff
@@ -177,6 +177,81 @@ def email_addresses_of_type_without_notification_sent(type)
...
+    email
```
Blaidd-Drwg: This could be made less verbose by passing the variables as an argument to `new`, like so
```ruby
Email.new(
  hide_recipients: false,
  stuff: more_stuff)
```

app/models/event.rb
```diff
@@ -177,6 +177,81 @@ def email_addresses_of_type_without_notification_sent(type)
...
+  end
+
+  # Returns all email addresses of user that fit the given criteria
+  #
+  # @param all Boolean
+  # @param groups [String]
+  # @param users [Int]
+  # @return [String] list of email addresses
```
Blaidd-Drwg: Missing parameter descriptions, formatting

app/models/event.rb
```diff
@@ -177,6 +177,81 @@ def email_addresses_of_type_without_notification_sent(type)
...
+    if all
+      email_addresses_of_accepted_applicants
+    else
+      (email_addresses_of_groups(groups) + email_addresses_of_users(users)).uniq.join(',')
+    end
+  end
+
+  # Returns a list of tuples containing all participant names and their id
+  #
+  # @return [[String, Int]]
```
Blaidd-Drwg: The format for that is `Array<Array<String, Int>>`, I think.

app/models/event.rb
```diff
@@ -177,6 +177,81 @@ def email_addresses_of_type_without_notification_sent(type)
...
+  def participants_with_id
+    participants.map{|participant| [participant.profile.first_name + ' ' + participant.profile.last_name, participant.id]}
```
Blaidd-Drwg: You can use `Profile#name` to get the full name

app/models/event.rb
```diff
@@ -177,6 +177,81 @@ def email_addresses_of_type_without_notification_sent(type)
...
+  end
```
Blaidd-Drwg: I don't know if this one-liner with such a rare use case justifies adding a method to the class's interface. Maybe move it to where it is called instead?

app/models/event.rb
```diff
@@ -177,6 +177,81 @@ def email_addresses_of_type_without_notification_sent(type)
...
+
+  # Returns a list of group names and their id
+  #
+  # @return []
+  def groups_with_id
+    existing = ParticipantGroup::GROUPS.select do |group_id,colorcode|
+      participant_groups.where(:group => group_id).count > 0
```
Blaidd-Drwg: Instead of counting: `participant_groups.where(:group => group_id).any?`

app/models/event.rb
```diff
@@ -177,6 +177,81 @@ def email_addresses_of_type_without_notification_sent(type)
...
+  # @return [[String, Int]]
+  def participants_with_id
+    participants.map{|participant| [participant.profile.first_name + ' ' + participant.profile.last_name, participant.id]}
+  end
+
+  # Returns a list of group names and their id
+  #
+  # @return []
+  def groups_with_id
+    existing = ParticipantGroup::GROUPS.select do |group_id,colorcode|
```
Blaidd-Drwg: You can designate a variable you don't need by naming it with an undescore, like `|group_id, _|`

app/models/event.rb
```diff
@@ -177,6 +177,81 @@ def email_addresses_of_type_without_notification_sent(type)
...
+  # Returns a list of tuples containing all participant names and their id
+  #
+  # @return [[String, Int]]
+  def participants_with_id
+    participants.map{|participant| [participant.profile.first_name + ' ' + participant.profile.last_name, participant.id]}
+  end
+
+  # Returns a list of group names and their id
+  #
+  # @return []
```
Blaidd-Drwg: Same as above

app/models/event.rb
```diff
@@ -177,6 +177,81 @@ def email_addresses_of_type_without_notification_sent(type)
...
+    existing.map do |group_id,colorcode|
+      [I18n.t("participant_groups.options.#{colorcode}"), group_id]
+    end
+  end
+
+  # Returns a list of email addresses of all participants that are in one of the given groups
+  #
+  # @return [String]
+  def email_addresses_of_groups(groups)
+    groups.reduce([]) {|addrs, group| addrs + email_addresses_of_group(group)}.uniq
```
Blaidd-Drwg: >addrs
You're not saving that many letters, you know. Might as well write it out :grin:

app/models/event.rb
```diff
@@ -177,6 +177,81 @@ def email_addresses_of_type_without_notification_sent(type)
...
+  end
+
+  def email_addresses_of_group(group)
+    participant_groups.where(:group => group).map {|participant_group| participant_group.user.email}
+  end
+
+  # Returns a list of email addresses
+  def email_addresses_of_users(users)
+    user = User.where('id in (:ids)', ids: users)
```
Blaidd-Drwg: I think you can just say `where(id: users)`, might wanna double check before you do it though :wink:

app/views/emails/_email_form.erb
```diff
@@ -1,4 +1,4 @@
-<%= form_for @email, url: {action: 'submit', controller: 'emails', status: params[:status]}, html: {class: 'email-form form-horizontal'} do |f| %>
+<%= form_for @email, url: @send_generic ? {action: 'submit_generic', controller: 'emails'} : {action: 'submit', controller: 'emails', status: params[:status]}, html: {class: 'email-form form-horizontal'} do |f| %>
```
Blaidd-Drwg: This line is pretty long, and using the ternary operator on a multi-line statement wouldn't be great either. Maybe set a variable in an if block before this line?

Blaidd-Drwg (changes requested):
I'll take a break here.
In general, I'm not a fan of having tons of one-line methods with exactly one use case, but I guess that comes down to preference. What I do think however is that most of those could be made private. That way you don't clutter the interface that much. Also, you have to write less documentation :stuck_out_tongue_winking_eye:

app/controllers/emails_controller.rb
Raandom: So that the template knows whether it should redirect to the submit or the submit_generic action. Wasn't necessary initially but they changed the email interface within the past weeks to make it far more specific then it was originally. Could use 2 different views instead, but thought it would be more convenient to just pass a parameter this way. I dont know whether there might be an better way (one could think of sending an path in there?)

app/controllers/emails_controller.rb
Raandom: There is no enabled draft sending for thoose emails, as the drafts always refer to an acceptance or rejection email, which is not the case here (and if I'm honest, I was too lazy to also adapt the template part of the mail system). But on the other hand I could disable the functionality in the view then. Mh. (This would probably also remove the need to provide this strange parameter to the view...)

app/controllers/emails_controller.rb
Blaidd-Drwg: Those are all valid considerations, but what I meant is: why do you set it here in this method? :smile: 
`submit` is called by the view and doesn't render anything....
Oh wait, I understood while I was writing this. `submit` calls `send_email`, which could render `email`. Maybe you could make it clearer that the variable is for that purpose by moving the assignment to `send_email`, just before the call to `render`?

app/views/emails/_email_form.erb
```diff
@@ -45,7 +46,9 @@
     </div>
   </div>
   <div class="btn-group pull-right" role="group">
+    <% unless @send_generic %>
     <%= f.submit t('.save_template'), name: 'save', class: 'email-submit btn btn-default' %>
```
Blaidd-Drwg: Indentation

app/views/emails/email.erb
```diff
@@ -1,11 +1,14 @@
...
 </div>
 <div>
-  <div class="email-form col-md-8">
+  <div class="email-form col-md-<%= width %>">
     <%= render partial: 'emails/email_form', locals: { email: @email } %>
   </div>
+  <% unless @send_generic %>
   <div class="email-templates col-md-4">
     <%= render partial: 'emails/templates', locals: { template: @templates } %>
   </div>
```
Blaidd-Drwg: Indentation

app/views/events/_email_modal.html.erb
```diff
@@ -0,0 +1,34 @@
...
+              <%= select_tag(:groups, options_for_select(@event.groups_with_id), class:'form-control', multiple: true) %>
+            </div>
+            <div class="form-group">
+              <%= label_tag(:all, t('.select_all')) %>
+              <%= check_box_tag(:all, '1', false, class: 'form-control') %>
+            </div>
+            <div class="form-group">
+              <%= label_tag(:sender, t('.sender')) %>
+              <%= text_field_tag(:sender, '', class: 'form-control') %>
+            </div>
```
Blaidd-Drwg: This isn't needed, the sender can be set in the email form

app/views/events/_email_modal.html.erb
```diff
@@ -0,0 +1,34 @@
...
+              <%= select_tag(:users, options_for_select(@event.participants_with_id), class:'form-control', multiple: true) %>
+            </div>
+            <div class="form-group">
+              <%= label_tag(:groups, t('.select_groups')) %>
+              <%= select_tag(:groups, options_for_select(@event.groups_with_id), class:'form-control', multiple: true) %>
+            </div>
+            <div class="form-group">
+              <%= label_tag(:all, t('.select_all')) %>
+              <%= check_box_tag(:all, '1', false, class: 'form-control') %>
+            </div>
```
Blaidd-Drwg: The "all" checkbox makes more sense at the top of the form, IMO.

Blaidd-Drwg: Currently the UI is a little confusing.
1) The multiselect is not intuitive, an average user won't try ctrl+click. Possible solutions are either change the behavior of the multiselect to work with normal clicks (my favorite), or add a tooltip that explains how to use it (probably easier to implement, less comfortable to use).
2) There are currently two ways to select the recipients: either with the multiselects or with the "all" checkbox, and checking the latter overrides the selections in the multiselect. Maybe you could disable the two multiselects when the checkbox is checked? Or have the checkbox automatically select every user in the multiselects. Either way, it should be just a few lines of Javascript.

app/models/event.rb
```diff
@@ -177,6 +177,84 @@ def email_addresses_of_type_without_notification_sent(type)
...
+  def participants_with_id
+    participants.map{|participant| [participant.profile.name, participant.id]}
+  end
+
+  # Returns a list of group names and their id
+  #
+  # @return Array<Array<String, Int>>
+  def groups_with_id
+    existing = ParticipantGroup::GROUPS.select do |group_id,_|
+      participant_groups.where(:group => group_id).any?
```
Blaidd-Drwg: Please also check that `group_id` is not 0, that is the id for no group.

config/locales/de.events.yml
```diff
@@ -85,6 +86,14 @@ de:
     type:
       private: "Privat"
       public: "Öffentlich"
+    email_modal:
+      send_email: "E-Mail an Teilnehmer senden"
+      select_users: "Nutzer auswählen"
+      select_groups: "Gruppen auswählen"
+      select_all: "Alle"
+      sender: "Absenderadresse"
+      send_email_label: "Absenden"
```
Blaidd-Drwg: Confusing, since it doesn't actually send the emails. "Weiter" will do.

config/locales/de.events.yml
```diff
@@ -85,6 +86,14 @@ de:
     type:
       private: "Privat"
       public: "Öffentlich"
+    email_modal:
+      send_email: "E-Mail an Teilnehmer senden"
+      select_users: "Nutzer auswählen"
+      select_groups: "Gruppen auswählen"
+      select_all: "Alle"
```
Blaidd-Drwg: "Alle auswählen" would be clearer

spec/controllers/events_controller_spec.rb
```diff
@@ -157,6 +157,17 @@
...
     end
 
+    describe "GET #send_participants_email" do
+      before :each do
+        @user = FactoryGirl.create(:user_with_profile, role: :admin)
+        sign_in @user
+      end
+      it "should assign a new email to accepted applications as @email" do
+        get :send_participants_email, id: @event.to_param, session: valid_session, all: true, groups: [], users:[], sender: 'workshop.portal@hpi.de'
+        expect(assigns(:email)).to have_attributes(hide_recipients: false, recipients: @event.email_addresses_of_accepted_applicants, reply_to: 'workshop.portal@hpi.de', subject: '', content: '')
```
Blaidd-Drwg: Please avoid hardcoding the email addresses. Use `Rails.configuration.reply_to_address` instead.

spec/features/participants_spec.rb
```diff
@@ -36,36 +36,37 @@
     link_name = I18n.t("activerecord.attributes.participant_group.group")
     click_link link_name
     sorted_by_group = @event.participants.sort_by {|p| @event.participant_group_for(p).group }
-    names = sorted_by_group.map {|p| p.profile.name }
+    names = sorted_by_group.map {|p| (p.email)}
     expect(page).to contain_ordered(names)
 
+    expect(page).to have_link(link_name)
```
Blaidd-Drwg: Does this fix an issue with Turbolinks or why is this needed?

spec/features/participants_spec.rb
```diff
@@ -79,6 +80,44 @@
...
+    end
+    visit event_participants_path(@event)
+    check :all
+    fill_in(:sender, with: 'test@test.com')
+    click_button(I18n.t('events.email_modal.send_email_label'))
+    fill_in('email_subject', with: 'Subject')
+    fill_in('email_content', with: 'Content')
+    expect{click_button I18n.t('emails.email_form.send')}.to change{ActionMailer::Base.deliveries.count}.by(1)
+    expect(page).to have_text(I18n.t('emails.submit.sending_successful'))
+  end
```
Blaidd-Drwg: You don't check whether all participants were actually selected.

spec/features/participants_spec.rb
```diff
@@ -79,6 +80,44 @@
...
+  scenario "logged in as Organizer I cant send an invalid email to all participants" do
+    login(:organizer)
+    for i in 1..5
+      user = FactoryGirl.create(:user)
+      profile = FactoryGirl.create(:profile, user: user, last_name: i.to_s)
+      application_letter = FactoryGirl.create(:application_letter_accepted, user: user, event: @event)
+      participant_group = FactoryGirl.create(:participant_group, user: user, event: @event, group: i)
+    end
+    visit event_participants_path(@event)
+    check :all
```
Blaidd-Drwg: Since you use the same code to initialize both tests, you can move it to a separate function.

spec/models/event_spec.rb
```diff
@@ -290,4 +290,56 @@
       expect(event.participant_selection_locked).to eq(acceptances_sent || rejections_sent)
     end
   end
+
+  it "computes the email addresses of all participants" do
+    event = FactoryGirl.create(:event)
```
Blaidd-Drwg: These tests have some looong lines, please add line breaks where needed.

spec/models/event_spec.rb
```diff
@@ -290,4 +290,56 @@
...
+    accepted_application_letter_1 = FactoryGirl.create(:application_letter_accepted, :event => event, :user => FactoryGirl.create(:user, :profile => FactoryGirl.create(:profile)))
+    accepted_application_letter_2 = FactoryGirl.create(:application_letter_accepted, :event => event, :user => FactoryGirl.create(:user, :profile => FactoryGirl.create(:profile)))
+    accepted_application_letter_3 = FactoryGirl.create(:application_letter_accepted, :event => event, :user => FactoryGirl.create(:user, :profile => FactoryGirl.create(:profile)))
```
Blaidd-Drwg: The factory for ApplicationLetter already creates a user with profile.

spec/models/event_spec.rb
```diff
@@ -290,4 +290,56 @@
...
+    rejected_application_letter = FactoryGirl.create(:application_letter_rejected, :event => event, :user => FactoryGirl.create(:user))
+    [accepted_application_letter_1, accepted_application_letter_2, accepted_application_letter_3, rejected_application_letter].each { |letter| event.application_letters.push(letter) }
```
Blaidd-Drwg: This line isn't needed, Rails adds the automatically.

spec/models/event_spec.rb
```diff
@@ -290,4 +290,56 @@
...
+    expect(event.email_addresses_of_participants(true, [], [])).to include(accepted_application_letter_1.user.email, accepted_application_letter_2.user.email, accepted_application_letter_3.user.email)
+  end
+
+  it "computes the email addresses of a group" do
+    event = FactoryGirl.create(:event)
+    accepted_application_letter_1 = FactoryGirl.create(:application_letter_accepted, :event => event, :user => FactoryGirl.create(:user, :profile => FactoryGirl.create(:profile)))
+    accepted_application_letter_2 = FactoryGirl.create(:application_letter_accepted, :event => event, :user => FactoryGirl.create(:user, :profile => FactoryGirl.create(:profile)))
+    accepted_application_letter_3 = FactoryGirl.create(:application_letter_accepted, :event => event, :user => FactoryGirl.create(:user, :profile => FactoryGirl.create(:profile)))
+    rejected_application_letter = FactoryGirl.create(:application_letter_rejected, :event => event, :user => FactoryGirl.create(:user))
```
Blaidd-Drwg: Again, all these identical initializations can be moved to a separate function.

spec/models/event_spec.rb
```diff
@@ -290,4 +290,56 @@
...
+    [accepted_application_letter_1, accepted_application_letter_2, accepted_application_letter_3, rejected_application_letter].each { |letter| event.application_letters.push(letter) }
+    expect(event.participants_with_id).to include(
+        [accepted_application_letter_1.user.profile.first_name + ' ' + accepted_application_letter_1.user.profile.last_name, accepted_application_letter_1.user.id],
+        [accepted_application_letter_2.user.profile.first_name + ' ' + accepted_application_letter_2.user.profile.last_name, accepted_application_letter_2.user.id],
+        [accepted_application_letter_3.user.profile.first_name + ' ' + accepted_application_letter_3.user.profile.last_name, accepted_application_letter_3.user.id],
```
Blaidd-Drwg: You can use `Profile#name` here

spec/models/event_spec.rb
```diff
@@ -290,4 +290,56 @@
...
+    rejected_application_letter = FactoryGirl.create(:application_letter_rejected, :event => event, :user => FactoryGirl.create(:user))
+    participant_group1 = FactoryGirl.create(:participant_group, :event => event, :user => accepted_application_letter_1.user, :group => 2)
+    [accepted_application_letter_1, accepted_application_letter_2, accepted_application_letter_3, rejected_application_letter].each { |letter| event.application_letters.push(letter) }
+    result = event.email_addresses_of_participants(false, [participant_group1.group], [])
+    expect(result).to include(accepted_application_letter_1.user.email)
+    expect(result).to_not include(accepted_application_letter_2.user.email, accepted_application_letter_3.user.email)
+  end
+
+
+  it "computes the email addresses of certain participants" do
```
Blaidd-Drwg: Please also test what happens when both `groups` and `users` are non-empty in email_addresses_of_participants

spec/views/events/participants.html.erb_spec.rb
```diff
@@ -113,4 +113,16 @@
...
+  it "displays an modal that allows selection of email target" do
+    render
+    expect(rendered).to have_css('div#send_participant_email_modal')
+  end
+
+  it "allows selection of participants in the modal" do
+    application_letter = FactoryGirl.create(:application_letter, user: FactoryGirl.create(:user_with_profile, role: :admin), event: @event, status: 1)
+    @event.application_letters.push(application_letter)
+    render
+    expect(rendered).to have_select('users', :with_options => [application_letter.user.profile.first_name + ' ' + application_letter.user.profile.last_name])
```
Blaidd-Drwg: `Profile#name`

spec/views/events/participants.html.erb_spec.rb
```diff
@@ -113,4 +113,16 @@
...
     expect(rendered).to have_css('input#print_participant_list')
   end
+
+  it "displays an modal that allows selection of email target" do
+    render
+    expect(rendered).to have_css('div#send_participant_email_modal')
+  end
+
+  it "allows selection of participants in the modal" do
+    application_letter = FactoryGirl.create(:application_letter, user: FactoryGirl.create(:user_with_profile, role: :admin), event: @event, status: 1)
```
Blaidd-Drwg: Why does the user have to be an admin?
You can use the `accepted_application_letter` factory instead of setting the status.

spec/views/events/show.html.erb_spec.rb
```diff
@@ -3,8 +3,7 @@
 RSpec.describe "events/show", type: :view do
   before(:each) do
     @event = assign(:event, FactoryGirl.create(:event, :with_two_date_ranges))
-    @application_letter = FactoryGirl.create(:application_letter, user: FactoryGirl.create(:user, role: :admin), event: @event)
-    @application_letter.user.profile = FactoryGirl.build(:profile)
+    @application_letter = FactoryGirl.create(:application_letter, user: FactoryGirl.create(:user_with_profile, role: :admin), event: @event)
     @event.application_letters.push(@application_letter)
```
Blaidd-Drwg: This isn't needed

spec/views/events/show.html.erb_spec.rb
```diff
@@ -3,8 +3,7 @@
 RSpec.describe "events/show", type: :view do
   before(:each) do
     @event = assign(:event, FactoryGirl.create(:event, :with_two_date_ranges))
-    @application_letter = FactoryGirl.create(:application_letter, user: FactoryGirl.create(:user, role: :admin), event: @event)
-    @application_letter.user.profile = FactoryGirl.build(:profile)
+    @application_letter = FactoryGirl.create(:application_letter, user: FactoryGirl.create(:user_with_profile, role: :admin), event: @event)
```
Blaidd-Drwg: Admin rights?

Blaidd-Drwg (changes requested):

app/controllers/emails_controller.rb
```diff
@@ -10,23 +10,33 @@ def show
...
     @email = Email.new(hide_recipients: true, reply_to: Rails.configuration.reply_to_address, recipients: @addresses.join(','),
                        subject: '', content: '')
-
+    @send_generic = false
     render :email
   end
 
-  def submit
+  def submit_application_result
+    @send_generic = false
```
Blaidd-Drwg: Can you move this before the call to `render` as well?

app/models/event.rb
```diff
@@ -177,6 +177,83 @@ def email_addresses_of_type_without_notification_sent(type)
...
+  # @return [String] Concatenation of all email addresses of accepted applications, seperated by ','
+  def email_addresses_of_accepted_applicants
+    participants.collect { |user| user.email }.join(',')
+  end
+
+  # Returns a new email to a set of participants
+  #
+  # @param all Boolean If set to true, addresses all participants
+  # @param groups Array<Integer> The group-ids whoose members should be addressed
+  # @param users Array<Integer> The user-ids which should be addressed
```
Blaidd-Drwg: The type should be in square brackets: `@param all [Boolean] If set to true, addresses all participants` and so on

app/models/event.rb
```diff
@@ -177,6 +177,83 @@ def email_addresses_of_type_without_notification_sent(type)
...
+    existing.map do |group_id,color_code|
+      [I18n.t("participant_groups.options.#{color_code}"), group_id]
+    end
+  end
+
+  # Returns a list of email addresses of all participants that are in one of the given groups
+  #
+  # @return [String]
+  def email_addresses_of_groups(groups)
+    groups.reduce([]) {|addresses, group| addresses + email_addresses_of_group(group)}.uniq
```
Blaidd-Drwg: This breaks when no group is selected in the frontend, then `groups` is `nil`.

app/models/event.rb
Blaidd-Drwg: ?

app/models/event.rb
```diff
@@ -177,6 +177,83 @@ def email_addresses_of_type_without_notification_sent(type)
...
+  # @param groups Array<Integer> The group-ids whoose members should be addressed
+  # @param users Array<Integer> The user-ids which should be addressed
+  # @return Email new email
+  def generate_participants_email(all, groups, users)
+    Email.new(
+        :hide_recipients => false,
+        :recipients => email_addresses_of_participants(all, groups, users),
+        :reply_to => '',
+        :subject => '',
+        :content => ''
```
Blaidd-Drwg, line 192: Is setting the empty strings needed?

app/views/events/_email_modal.html.erb
```diff
@@ -0,0 +1,30 @@
...
+            <div class="form-group">
+              <%= label_tag(:users, t('.select_users')) %>
+              <%= select_tag(:users, options_for_select(@event.participants_with_id), class:'form-control', multiple: true) %>
+            </div>
+            <div class="form-group">
+              <%= label_tag(:groups, t('.select_groups')) %>
+              <%= select_tag(:groups, options_for_select(@event.groups_with_id), class:'form-control', multiple: true) %>
+            </div>
```
Blaidd-Drwg, line 44: As suspected, Turbolinks breaks this. It doesn't display correctly when the participants page is reached by clicking a link.

app/models/event.rb
```diff
@@ -177,6 +177,83 @@ def email_addresses_of_type_without_notification_sent(type)
...
+  def participants_with_id
+    participants.map{|participant| [participant.profile.name, participant.id]}
+  end
+
+  # Returns a list of group names and their id
+  #
+  # @return Array<Array<String, Int>>
+  def groups_with_id
+    existing = ParticipantGroup::GROUPS.select do |group_id,_|
+      group_id != 0 and participant_groups.where(:group => group_id).any?
```
Blaidd-Drwg: The style guide suggests always using `&&` instead of `and` because of weird operator precedence.

Blaidd-Drwg (changes requested):

radscheit (approved):

Blaidd-Drwg (approved):

PR #343, opened 2017-01-18
286 3.28 Deny application for coaches and organizers

Lando-L:
Removed the 'apply' button for coaches and organizers for all events


app/models/ability.rb
```diff
@@ -43,6 +43,8 @@ def initialize(user)
       # Pupils can upload their letters of agreement
       can [:create], AgreementLetter
       can [:new, :create], Request
+      # Pupils can apply for events
```
frederike-ramin: Comment is pretty unnecessary.

app/models/ability.rb
```diff
@@ -51,6 +53,8 @@ def initialize(user)
       can [:print_applications], Event
       can :manage, Request
       cannot [:check], ApplicationLetter
+      # Coaches can not apply for events
```
frederike-ramin: same here.

app/models/ability.rb
```diff
@@ -59,6 +63,8 @@ def initialize(user)
       # Organizers can view, edit and print Applications, view participants for, view and upload materials for, print agreement letters for and manage Events
       can [:view_applicants, :edit_applicants, :view_participants, :print_applications, :manage, :view_material, :upload_material, :print_agreement_letters], Event
       can :manage, Request
+      # Coaches can not apply for events
```
frederike-ramin: See the pattern? ;) (also - organizers, not coaches.)

spec/views/events/_event.html.erb_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    it "as #{role} you can see the apply button" do
+      user = FactoryGirl.create(:user, role: role)
+      sign_in user
+      render partial: 'events/event.html.erb', locals: {event: @event}
+      expect(rendered).to have_link(t('helpers.links.apply'), href: new_application_letter_path(event_id: @event.id))
+    end
+  end
+
+  %i[coach organizer].each do |role|
+    it "as #{role} you can not see the apply button" do
```
frederike-ramin: cannot^^

spec/views/events/_show_event.html.erb_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    it "as #{role} you can see the apply button" do
+      user = FactoryGirl.create(:user, role: role)
+      sign_in user
+      render
+      expect(rendered).to have_link(t('helpers.links.apply'), href: new_application_letter_path(event_id: @event.id))
+    end
+  end
+
+  %i[coach organizer].each do |role|
+    it "as #{role} you can not see the apply button" do
```
frederike-ramin: cannot. Yeyy - copy, paste. :D

frederike-ramin (changes requested):
Please write ability tests too.

frederike-ramin (dismissed):

app/models/ability.rb
```diff
@@ -59,6 +61,7 @@ def initialize(user)
       # Organizers can view, edit and print Applications, view participants for, view and upload materials for, print agreement letters for and manage Events
       can [:view_applicants, :edit_applicants, :view_participants, :print_applications, :manage, :view_material, :upload_material, :print_agreement_letters], Event
       can :manage, Request
+      cannot [:apply], Event
```
aloeser: I think we dont need this line, it should already be inherited from coach access rights

spec/models/ability_spec.rb
```diff
@@ -258,4 +258,18 @@
...
   end
+
+  it "can apply for events as pupil" do
+    user = FactoryGirl.create(:user, role: :pupil)
+    ability = Ability.new(user)
+    expect(ability).to be_able_to(:apply, Event)
+  end
+
+  %i[coach organizer].each do |role|
+    it "can apply for events as #{role}" do
```
aloeser: cannot, not can ;)

aloeser (changes requested):

app/models/ability.rb
Lando-L: The test fails without this line

app/models/ability.rb
```diff
@@ -60,6 +62,7 @@ def initialize(user)
       can [:view_applicants, :edit_applicants, :view_participants, :print_applications, :manage, :view_material, :upload_material, :print_agreement_letters, :download_material, :view_unpublished], Event
       can :send_email, Email
       can :manage, Request
+      cannot :apply, Event
```
aloeser: I'm still not sure if we need this line. It should be already inherited from coach.
The problem is the :manage in line 60/62, since :manage means an organizer can do everything with an event.
If this should be in there, I guess we could skip all the other permissions for event

aloeser (changes requested):

app/models/ability.rb
```diff
@@ -51,13 +52,14 @@ def initialize(user)
...
       cannot :check, ApplicationLetter
     end
     if user.role? :organizer
       can [:index, :show], Profile
       can [:index, :show, :view_and_add_notes, :update_status], ApplicationLetter
       cannot :update, ApplicationLetter
-      can [:view_applicants, :edit_applicants, :view_participants, :print_applications, :manage, :view_material, :upload_material, :print_agreement_letters, :download_material, :view_unpublished], Event
+      can [:view_applicants, :edit_applicants, :view_participants, :print_applications, :view_material, :upload_material, :print_agreement_letters, :download_material, :view_unpublished, :new, :edit, :destroy], Event
```
aloeser: You still need to add an extra permission like :view_custom_application_fields, and update the ability spec accordingly.
Further, I found two usages of :manage, Event. I guess those are the one that cause your tests to fail atm.
Use your editor or execute something like grep "can? :manage, Event -r app" to find them

aloeser (changes requested):

spec/models/ability_spec.rb
```diff
@@ -271,10 +271,18 @@
...
+  it "can new, edit and destroy events as organzier" do
     user = FactoryGirl.create(:user, role: :organizer)
     ability = Ability.new(user)
-    expect(ability).to be_able_to(:manage, Event)
+    expect(ability).to be_able_to(:new, Event)
+    expect(ability).to be_able_to(:edit, Event)
+    expect(ability).to be_able_to(:destroy, Event)
+  end
+
+  it "can view custom application fields as organizer" do
```
frederike-ramin: How about a view test for that?

frederike-ramin (dismissed):

app/views/events/_event.html.erb
```diff
@@ -28,7 +28,9 @@
       </div>
 
       <div class="pull-right">
-        <%= render :partial => 'events/apply_button', :locals => {event: event} %>
+        <% if can? :apply, Event %>
+            <%= render :partial => 'events/apply_button', :locals => {event: event} %>
```
querenker: To reduce code duplication you can put this check inside the apply_button partial. I think currently you even forgot one position to add this test.

spec/views/events/_apply_button_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'rails_helper'
+
+RSpec.describe "events/_apply_button", type: :view do
+  before(:each) do
+    @event = assign(:event, FactoryGirl.create(:event))
```
aloeser: I'd suggest to use the trait :in_application_phase.
I know that an event is per default in application phase, but if this changes, those tests loose some of their value

aloeser (changes requested):

spec/views/events/_apply_button_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'rails_helper'
+
+RSpec.describe "events/_apply_button", type: :view do
+  before(:each) do
+    @event = assign(:in_application_phase, FactoryGirl.create(:event))
```
aloeser: I think you misunderstood me.
It was `assign(:event, FactoryGirl.create(:event))` before, and what i meant was something like `assign(:event,  FactoryGirl.create(:event, :in_application_phase)`, in order to give the event another property. The way it is now, the instance var in_application_phase is set, and thats not what we want :)
Also, I missed at the first look that this is a test for a partial. And since the @event is directly passed to the partial in the render call, we can skip the assign()-part.
So, after all this line should be `@event = FactoryGirl.create(:event, :in_application_phase)`

aloeser (changes requested):

aloeser (approved):

ekrebs5 (approved):

Issue #321, opened 2017-01-12
Custom variable application fields in application


bjrne:
- [ ] render fields on application fill-out page
**Dependency:**
#322

Helene0:
**As**
user

**I want to**
be able to fill out the custom fields that the event organizer provided

**In order to**
share the information that the organizers need to pick me

MasterCarl:
🍪 **20**

PR #354, opened 2017-01-18
Application letter form with variable fields

MasterCarl:
References #321


db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170118185870) do
+ActiveRecord::Schema.define(version: 20170118192270) do
```
bjrne: Since there is no migration, there is not need for this.

bjrne (approved):
Works locally and tests seem appropriate. Please remove the schema change though, it's not necessary and potentially introduces conflicts.

PR #355, opened 2017-01-18
312 email after request

bjrne:
This insanely complex PR relies on the idea that mails don't need to be tested since we use a framework for them. Please cry if I still need to test our wrapper function. Oh yeah and I forgot to rename the test in the last PR.


app/controllers/requests_controller.rb
```diff
@@ -23,8 +23,9 @@ def edit
   # POST /requests
   def create
     @request = Request.new(request_params)
-
     if @request.save
+      Mailer.send_generic_email(false, @request.email, 'workshop.portal@gmail.com', I18n.t('requests.email.topic'),
```
cmfcmf: `workshop.portal@gmail.com` should probably be placed in some kind of configuration file.

aloeser: I think you should add a test for this line. Not for testing the mailer functionality, but rather for testing that a mail is sent. Otherwise one could delete this line, and the tests would be green, allthough no email was sent.
Maybe something like 
```
it "sends an email" do
  expect{ post :create, request: valid_attributes, session: valid_session }.to change{ActionMailer::Base.deliveries.count}.by(1)
end
```

bjrne: @cmfcmf That was my idea too, but after talking to @Baschdl's team I found out they don't really have a place for that yet and hardcode it every time. I was hoping for good input on this  in the PR here :) Do you have an idea where to put it? I guess it should be changeable in and organizer/admin view later.

bjrne: @aloeser sounds good. Maybe someone who has more experience with mails can give your code a look? @TimBasel

timbasel: The problem is that the @gmail address was planed as just an internal email address for sending. So there was no intention to ever have to change it. 
But in this case it would be a good idea to make the reply_to address configurable by organizers/admins

bjrne: @TimBasel but where would you save it?

cmfcmf: @bjrne Take a look at how the "unreasonably long event time span" is stored: https://github.com/hpi-swt2/workshop-portal/blob/7d76af4af777d74b8917659ad7e9d6d3db58ab83/config/application.rb#L30-L30

bjrne: Thanks @cmfcmf, Tom already showed me. That's why it's already implemented on awaits a review ;)

app/mailers/application_mailer.rb
```diff
@@ -1,4 +1,4 @@
 class ApplicationMailer < ActionMailer::Base
-  default from: "workshop.portal@gmail.com"
+  default from: Rails.configuration.default_reply_to_email
```
aloeser: This change is wrong. We actually send emails with this adress.
I agree that it might be good to read this address from the config/environment too, but from and reply-to are two different emails.
I dont know the correct var atm, I'll have a look later on

config/application.rb
```diff
@@ -28,5 +28,8 @@ class ApplicationLetter < Rails::Application
 
     # see Event::unreasonably_long, unit in days
     config.unreasonably_long_event_time_span = 300
+
+    # default reply_to email adress
+    config.default_reply_to_email = 'workshop-portal@gmail.com'
```
aloeser: If the system is supposed to run at/for the HPI, they probably prefer an @hpi.de adress as default reply-to?
@tobiduer?

config/locales/de.requests.yml
```diff
@@ -21,6 +21,10 @@ de:
       was_updated: "Die Anfrage wurde aktualisiert"
       was_accepted: "Die Anfrage wurde angenommen"
       was_deleted: "Die Anfrage wurde gelöscht"
+    email:
+      topic: "Ihre Anfrage auf dem HPI Workshop-Portal"
+      content: "Hallo!
+                Ihre Anfrage ist bei uns eingegangen und wir melden uns schnellstmöglich bei Ihnen um weitere Details abzustimmen."
```
aloeser: comma after 'Ihnen' missing

spec/controllers/requests_controller_spec.rb
```diff
@@ -182,6 +182,10 @@
         expect(assigns(:request)).to be_persisted
       end
 
+      it "sends an email" do
+        expect{ post :create, request: valid_attributes, session: valid_session }.to change{ ActionMailer::Base.deliveries.count }.by(1)
+      end
+
```
aloeser, line 194: I would suggest to add a similar test, that no email is sent, when using invalid params

spec/mailers/portal_mailer_spec.rb
```diff
@@ -22,7 +22,7 @@
     end
 
     it 'renders the sender email' do
-      expect(mail.from).to eq(['workshop.portal@gmail.com'])
+      expect(mail.from).to eq([Rails.configuration.default_reply_to_email])
```
aloeser: Same as the top comment, the from is actually exactly the senders adress, not the reply to, so we need to revert this change.
Moreover, I think "sets" would fit better than "renders" (not your change tho)

aloeser (changes requested):
At the moment, we send emails from workshop.portal@gmail.com. However, the PO's want another email to reply to. I guess that caused some confusion.
As far I could see, you found all occurences of the email :)

app/mailers/application_mailer.rb
aloeser: Tested it: You can still send emails, but the from-adress is still the correct one. So I guess gmail just ignores any from-fields that are not valid

aloeser: Right now, we can access the sender adress with ENV['gmail_username'].
It returns just the name before the @ tho

tobiasduerschmid:
@aloeser: You can leave this email address and document the place where to change it in the wiki in a section on how to setup the portal

cmfcmf:
@bjrne please ping me again once all other reviews of this PR have passed. Thanks!

cmfcmf:
Part of this PR is obsolete (sorry!) [now that I made email sending great again](https://github.com/hpi-swt2/workshop-portal/pull/487) and introduced such application configuration options myself.

aloeser (approved):

cmfcmf (approved):
Looks good now!

PR #356, opened 2017-01-19
Remove line breaks from sample data

bjrne:
In order to be able to see the real layout and not some weird line breaks when testing the platform.


cmfcmf (approved):
You could've also considered doing it like this: http://stackoverflow.com/a/10524123/2560557
But it's fine either way.

bjrne:
I didn't know that @cmfcmf. Id still tend to this version, because people can turn soft wraps on in their editor. With the other one I'd have to decide for everyone how long the lines should be. But very interesting for other cases!

PR #360, opened 2017-01-19
Prefix error page css so that it doesn't conflict. Fixes #284

### LINKED ISSUE #284: Every text centered after error
After this error:
image
if you go back via the back button in the browser every text is centered on every page and doesn't change (only after reload with Ctrl + F5):
image

tom95:
Now, I don't actually know whether this did anything because I was not able to deploy rails in production mode, so I didn't get to see the error messages, but this is a very strong suspicion.


bjrne (approved):
Nice idea, let's find out. Shouldn't ruin too much 🌳

PR #361, opened 2017-01-19
Block users from creating two profiles. Fixes #325

### LINKED ISSUE #325: Users are able to create multiple profiles
Steps to reproduce:
- Create a profile at `/profiles/new`
- Create another profile at `/profiles/new`

@hpi-swt2/scrum-master-2016

tom95:


bjrne (commented):
I tried it this way as well. I'd still add a check for when a user manually opens `/profiles/new` In this scenario, he won't see any hint, still fill out the profile only to get redirected to the original one without notice.

Regardless of the route I'd add a little alert or notice that explains why nothing happens.

tom95:
Good point. I changed it so that it redirects from the /new page as well. I don't think we need a hint here, because a user shouldn't even be aware of the fact that it is possible to create multiple profiles. He should just be demonstrated that he already has a profile if he is mistakenly brought to that page.

bjrne (approved):
Nice. I guess the PO has the last word on this, but I understand what you mean 🌊

PR #365, opened 2017-01-20
Fix error text color, fix count on error message, fix layout on requests

tom95:
The color scheme branch set the text color of all errors to white, hoping that they would always have a red background, which was not the case. This reverses to the default bootstrap error colors.

Furthermore, this makes the global error partial a little more smart, by counting the number of messages, instead of errors, which corresponds to the number of fields errors appeared in instead of the global number of errors. (So no more "In 9 Feldern sind Fehler" while you actually only have 8 fields).

And lastly, an error span on the requests page was incorrectly placed, breaking the layout of the btn-group.


bjrne (approved):
👍

PR #367, opened 2017-01-20
Set mail host in config. Fixes #299.

### LINKED ISSUE #299: Forgot Password functionality does not work
If you try to use the forgot password functionality, after sending the form, the page errors out. 

@hpi-swt2/scrum-master-2016

tom95:
More untested production fixes! The heroku logs complained about `default_url_options[:host]` not being set. This sets it, based on an env var that we'd need to add on heroku once this is up.


bjrne (approved):
Let's try this, nobody complained until now :man_artist:

PR #375, opened 2017-01-21
53 3.13 load email drafts

timbasel:
Load Emails Templates

- Click on Template sets the hide_recipients radio buttons, subject and content field with template content


spec/models/ability_spec.rb
```diff
@@ -302,4 +302,21 @@
     expect(ability).to be_able_to(:send_email, Email)
   end
 
+
+  %i[pupil coach].each do |role|
```
ekrebs5: These tests are already in the file, adding them would mean we have these tests 2-times

app/assets/javascripts/emails.js
```diff
@@ -13,4 +13,21 @@ jQuery(function() {
             }
         }
     });
-});
\ No newline at end of file
+});
+
+function loadTemplate(elem) {
```
Lando-L: Anstatt die Funktion ins Html reinzuschreiben würde ich mehr jQuery Funktionen nutzen und so vorgehen:
```
$('.email-template.list-group-item').click(function() {
  var subject = $(this).find('template-subject')[0].text()
  var content = ...
});
```

Lando-L (changes requested):

Lando-L (approved):

spec/models/ability_spec.rb
ekrebs5: Now as review: 
These tests are already in the file, adding them would mean we have these tests 2-times

ekrebs5 (dismissed):

frederike-ramin (dismissed):
You didn't add any test, right? One feature test could do the job.

Plus, I agree with @ekrebs5 about the doubled tests.

ekrebs5 (approved):

spec/factories/email_templates.rb
```diff
@@ -14,17 +14,17 @@
...
-    status :acceptance
-  end
+    trait :default do
+      status :default
+    end
 
-  factory :email_template_rejection, parent: :email_template do
-    status :rejection
-  end
+    trait :accepted do
```
aloeser: you should call it acceptance

spec/factories/email_templates.rb
```diff
@@ -14,17 +14,17 @@
...
+      status :acceptance
+    end
 
-  factory :email_template_default, parent: :email_template do
-    status :default
+    trait :rejected do
```
aloeser: you should call it rejection

aloeser (changes requested):
just some small factory trait name changes, for consistency reasons

spec/features/email_spec.rb
```diff
@@ -56,6 +56,18 @@
...
 
+  scenario "logged in as Organizer I can load an email template", js: true do
+    login(:organizer)
+    @template = FactoryGirl.create(:email_template, :acceptance)
+
+    visit event_email_show_path(@event, status: :acceptance)
+    first('.email-template').click
+
+    expect(page.find('#email_subject').value).to eq(@template.subject)
+    expect(page.find('#email_content').value).to eq(@template.content)
```
aloeser: please also check for hide_recipients

aloeser (changes requested):
Last small change, looks good otherwise :)

aloeser (approved):

Issue #244, opened 2017-01-05
US_3.27: UI for event states


tobiduer:
**As**
user of the portal

**I want to**
see exactly the UI elements applicable for the current state of the event

**in order to**
avoid errors and confusion

# Tasks

- [x] no apply buttons after application deadline

- [x] show application button

- [x] replace apply button with show application button after applying

- [x] radio buttons are viewable dependent on state (only in selection, or with what issue #242 does)

- [x] buttons for print all, accept all, sending emails and printing badges and viewing the participant list are viewable dependent on state

# Acceptance criteria

## Application phase

- When the deadline is over, no "Bewerben" buttons should be displayed (on any page!)

- When a user applied for an event the "Bewerben" buttons should be replace with "Bewerbung anzeigen" and link to the application checking page. (Refs #185)  

- When the the event is in the application state, then the selection of participants should not be possible.

- When the the event is in the application state, then there are not button for print all, accept all, sending emails and printing badges and viewing the participant list should be invisible. 

## Selection phase

- When the event is in the selection state, then the selection of participants should be possible (set application state to accepted, rejected, alternative but NOT to canceled)

- When the the event is in the selection state, then organizers can view the buttons for: print all, accept all, sending emails and but not for: printing badges and viewing the participant 

## Execution phase

- When the event is in the execution state, then organizers can view the buttons for: printing badges and viewing the participant but not for: print all, accept all, sending acceptance / rejection emails (expectation: Refs #243)
 
See Epic_3.3: Application States (Refs #233)

tobiduer:
@bjrne: Thank you :) I adjusted the US.

tobiasduerschmid:
is enhances by Refs #418

PR #376, opened 2017-01-21
244 3.27 buttons according to event states

ekrebs5:
This PR is part of #244 
It only focuses on the visibility of the buttons mentioned in the user story (Task: "buttons for print all, accept all, sending emails and printing badges and viewing the participant list are viewable dependent on state"). All the buttons are on the page of an event.

| button \ phase         | draft   | application | selection | execution |
|----------------------|----|-------------|-----------|-----------|
| print all                    |  no     | no          | yes       | no        |
| accept all               | no  | no          | yes       | no        |
| send accept/reject email | no   | no         | yes       | no        |
| print badges             | no     | no       | no        | yes       |
| participant list         | no     | no       | no        | yes       |

EDIT:
print badges is on the participant list site and its visibility is thus dependent on the participant list button


spec/views/events/show.html.erb_spec.rb
```diff
@@ -54,17 +54,8 @@
...
-
-  it "displays print badges button" do
-    render
-    expect(rendered).to have_link(t(:print_button_label, scope: 'events.badges'))
-  end
-
   it "should not display accept-all-button for non-organizers" do
+    @event = assign(:event, FactoryGirl.create(:event, :in_selection_phase))
     @event.max_participants = Float::INFINITY
     [:coach, :student].each do | each |
```
aloeser: not your change... but since when do we have :student as a role?

spec/views/events/show.html.erb_spec.rb
```diff
@@ -54,17 +54,8 @@
     expect(rendered).to have_link(t(:details, scope: 'events.applicants_overview'))
   end
 
-  it "displays print applications button" do
```
aloeser, line 57: why did you delete this test?

spec/views/events/show.html.erb_spec.rb
```diff
@@ -54,17 +54,8 @@
...
-    render
-    expect(rendered).to have_link(t(:print_all, scope: 'events.applicants_overview'))
-  end
-
-  it "displays print badges button" do
```
aloeser: why did you delete this test?

aloeser (commented):
It seems like feature tests for print badges and print application buttons are missing?
Also it might be a good idea to add a test, that when in application phase, even with pending applications or too many accepted applications the buttons are not visible. This seemed buggy for me, and it would be nice to have a test for it

aloeser:
@tobiduer should it be possible to see the event details page for an event in draft phase?
In that case, we would need some more tests

spec/views/events/show.html.erb_spec.rb
ekrebs5: this is now event phase dependent and covered in a later test
(as in this test would be duplicate with the exact lines)

spec/views/events/show.html.erb_spec.rb
ekrebs5: this is now event phase dependent and covered in a later test
(as in this test would be duplicate with the exact lines)

ekrebs5:
@aloeser 
"It seems like feature tests for print badges and print application buttons are missing?"
Could be, not our issue, so you should make a technical debt issue

"Also it might be a good idea to add a test, that when in application phase, even with pending applications or too many accepted applications the buttons are not visible. This seemed buggy for me, and it would be nice to have a test for it"
They are added now

spec/views/events/show.html.erb_spec.rb
ekrebs5: I don't know why that is there, should I change it to pupil? (looks like this was somehow not renamed, I think in the very beginning there was a student)

app/views/events/_applicants_overview.html.erb
```diff
@@ -57,36 +57,27 @@
...
         <%= link_to t('.accept_all'), event_accept_all_applicants_path(@event.to_param), :class => 'btn btn-default' %>
       <% end %>
-      <%= link_to t('.print_all'), print_applications_event_path(@event.to_param), :target => "_blank", :class => 'btn btn-default' %>
+      <% if @event.phase == :selection %>
+        <%= link_to t('.print_all'), print_applications_event_path(@event.to_param), :target => "_blank", :class => 'btn btn-default' %>
+      <% end %>
     </div>
 
-    <% if not @event.applications_classified? %>
-      <div class="btn-group pull-right tooltip-wrapper has-tooltip" role="group" data-toggle="tooltip" title="<%= t '.unclassified_applications_left'%>">
```
aloeser, line 67: you loose this line

app/views/events/_applicants_overview.html.erb
```diff
@@ -57,36 +57,27 @@
...
-        <%= link_to t('events.badges.print_button_label'), event_badges_path(@event), :class => 'btn btn-default'%>
-        <% if can? :send_email, Email %>
-          <%= button_tag t('.sending_acceptances'), type: 'button', class: 'send-emails-button btn btn-default', disabled: true %>
-          <%= button_tag t('.sending_rejections'), type: 'button', class: 'send-emails-button btn btn-default', disabled: true %>
-        <% end %>
-      </div>
-    <% elsif (@event.compute_free_places < 0) %>
-      <div class="btn-group pull-right tooltip-wrapper has-tooltip" role="group" data-toggle="tooltip" title="<%= t '.maximum_number_of_participants_exeeded'%>">
```
aloeser, line 74: you loose this line

spec/views/events/show.html.erb_spec.rb
```diff
@@ -101,4 +89,64 @@
...
+    render
+    expect(rendered).to_not have_link(t(:print_all, scope: 'events.applicants_overview'))
+    expect(rendered).to_not have_link(t(:accept_all, scope: 'events.applicants_overview'))
+    expect(rendered).to_not have_link(t(:sending_acceptances, scope: 'events.applicants_overview'))
+    expect(rendered).to_not have_link(t(:sending_rejections, scope: 'events.applicants_overview'))
+    expect(rendered).to_not have_link(t(:print_button_label, scope: 'events.badges'))
+    expect(rendered).to_not have_link(t(:show_participants, scope: 'events.participants'))
+  end
+
+  it "does not display the disabled send email buttons in application phase" do
```
aloeser: I suggest to append something like "despite unclassfied applications", to make clear what edge case you are testing.
Test itself looks good

aloeser (changes requested):
You need to give  app/views/events/_applicants_overview.html.erb another glance.
I think the two lines you deleted killed the tooltip on the send accept/reject mail buttons (which tell you that you cant send mails now, since too many participants/unclassified remain)

tobiasduerschmid:
@aloeser: Yes, but only for organizers

aloeser (approved):

spec/views/events/participants.html.erb_spec.rb
```diff
@@ -49,17 +49,42 @@
...
 
   it "disables the print badges button when there are no participants" do
+    @event = assign(:event, FactoryGirl.create(:event, :in_execution_phase))
     @participants = []
     render
     expect(rendered).to have_link(t(:print_button_label, scope: 'events.badges', disabled: true))
   end
 
+  it "does not show the print badges button when the event is in draft phase" do
+    @event = assign(:event, FactoryGirl.create(:event_with_accepted_applications, :in_draft_phase))
```
querenker, line 68: Is it even possible to have accepted application in the draft phase?

spec/views/events/show.html.erb_spec.rb
```diff
@@ -96,4 +89,78 @@
...
+    sign_in(FactoryGirl.create(:user, role: :organizer))
+    render
+    expect(rendered).to have_link(t(:print_all, scope: 'events.applicants_overview'))
+    expect(rendered).to have_link(t(:accept_all, scope: 'events.applicants_overview'))
+    expect(rendered).to have_link(t(:sending_acceptances, scope: 'events.applicants_overview'))
+    expect(rendered).to have_link(t(:sending_rejections, scope: 'events.applicants_overview'))
+    expect(rendered).to_not have_link(t(:show_participants, scope: 'events.participants'))
+  end
+
+  it "does display the disabled send email buttons in selection phase (when there are unclassified applications)" do
```
querenker: Just for consistency. In the other cases it is "it displays..." instead of "it does display...."

spec/views/events/participants.html.erb_spec.rb
ekrebs5: no, it isn't
I added new tests just now because the button was moved
is the participant page only visible when there are accepted participants?

spec/views/events/show.html.erb_spec.rb
```diff
@@ -96,4 +89,78 @@
...
+    @event = assign(:event, FactoryGirl.create(:event, :with_diverse_open_applications, :in_selection_phase))
+    sign_in(FactoryGirl.create(:user, role: :organizer))
+    render
+    expect(rendered).to have_button(t(:sending_acceptances, scope: 'events.applicants_overview'), disabled: true)
+    expect(rendered).to have_button(t(:sending_rejections, scope: 'events.applicants_overview'), disabled: true)
+  end
+
+  it "does display the disabled send email buttons in selection phase (when there are too many accepted applications)" do
```
querenker: same as above

querenker (approved):
Just some minor comments

spec/views/events/participants.html.erb_spec.rb
ekrebs5: wait, you mean like draft and selection are a subset of when they are no participants?

I think these tests should still be there, because in these cases there should be no button at all, while in the "normal" case with no participants it's disabled

PR #379, opened 2017-01-22
249 2.45 error message on passwort vergessen page

nstrelow:
references #249 

* replaced the error partial of all devise views with our own
* Enabled devise paranoid mode to avoid user enumeration
* Email required on password reset


philipp-bode (approved):
Seems sensible so far

Issue #286, opened 2017-01-10
US_3.28 Deny apply for coaches and organizers


tobiasduerschmid:
**As**
organizer or coach 

**I do not want to**
apply for camps

**in order to**
focus on my main activities. 

# Acceptance criteria

- on all pages, coaches and organizers never see apply buttons for events

PR #380, opened 2017-01-22
244 3.27 event state ui

querenker:
- if the deadline for an event is over the apply button should not displayed
- if a user already applied for an event a button that links to the check application page should be displayed (without apply button)
- radio buttons to change the application state should only be visible in selection phase


spec/views/events/show.html.erb_spec.rb
```diff
@@ -101,4 +101,35 @@
...
+  it "displays a button to view the application if application deadline if over for an event where the pupil has applied" do
+    pupil = FactoryGirl.create(:user, role: :pupil)
+    application_letter = FactoryGirl.create(:application_letter, user: pupil, event: @event)
+    @event.application_deadline = Date.yesterday
+    sign_in pupil
+    render
+    expect(rendered).to have_link(I18n.t("helpers.links.show_application"), href: check_application_letter_path(application_letter))
+  end
+
+  it "should not display radio buttons to change application statuses if the event is in application state" do
```
ekrebs5, line 111: this is a radio button test, are sure it should be in this pr? (I thought that it is separate task with a separate pr)

querenker: This functionality is already implemented but I thought a view test would be an good idea for this. If you like I can extract this test into an own PR?

ekrebs5: I don't think this is implemented yet? I tested it on your branch right now, I could see the radio buttons

I think "_applicants_overview.html.erb" should have that check, but at the moment it is only checking the locked boolean and editing rights.

It does make me wonder why this test works though, I can clearly see radio buttons in the application phase in my browser.

ekrebs5: (also radio buttons should probably be added to the pr description)

querenker: You are right there was something missing. Now it should be fixed :)

app/views/application_letters/_application_selective.html.erb
```diff
@@ -1,4 +1,4 @@
-<% if can? :edit_applicants, Event and not (application_letter.event.application_status_locked == true) %>
+<% if (can? :edit_applicants, Event) && (application_letter.event.phase == :selection) %>
```
ekrebs5, line 1: I think the "not (application_letter.event.application_status_locked == true)" should stay, since we want to lock it after sending emails

ekrebs5 (changes requested):

app/views/application_letters/_application_selective.html.erb
querenker: I think the selection of participants should only be available in the selection phase (therefore the name) and if the event is in the selection phase the application should not be locked, because otherwise it is not in the selection phase.

ekrebs5: Oh, I forgot that sending emails/locking puts the event in execution phase
Yes, you are correct

ekrebs5 (approved):

app/models/user.rb
```diff
@@ -55,6 +55,23 @@ def events
     accepted_applications.collect { |a| a.event }
   end
 
+  # Returns true iff. user has submitted an application letter for the given event
+  #
+  # @param [Event] given_event
+  # @return [Boolean]
+  def application_letter_for_event?(given_event)
+    return !self.application_letter_for_event(given_event).nil?
```
frederike-ramin, line 64: it would be more ruby-like to use "not" instead of "!", right?

app/views/events/_apply_button.erb
```diff
@@ -0,0 +1,7 @@
+<% if current_user.present? && current_user.application_letter_for_event?(event) %>
+    <%= link_to t("helpers.links.show_application"), check_application_letter_path(current_user.application_letter_for_event(event)),
```
frederike-ramin, line 2: There is also a application_letter show. Why won't you use that?

spec/features/events_spec.rb
```diff
@@ -118,25 +120,30 @@
...
   end
 
-  scenario "logged in as Organizer I can change application status with radio buttons" do
+  scenario "logged in as Organizer I can change application status with radio buttons in selection phase" do
     login(:organizer)
-    @event.application_status_locked = false
-    @event.save
     @pupil = FactoryGirl.create(:profile)
     @application_letter = FactoryGirl.create(:application_letter, event: @event, user: @pupil.user)
+    @event.application_status_locked = false
```
frederike-ramin: Maybe just creating another event in selection phase here would be more readable?

spec/views/events/index.html.erb_spec.rb
```diff
@@ -33,4 +33,25 @@
     expect(rendered).to have_link(href: edit_event_path(@event))
     expect(rendered).to have_link(href: event_path(@event), class: 'btn-danger')
   end
+
+  it "should not display apply button if application deadline is over" do
+    @event.application_deadline = Date.yesterday
+    assign(:events, [@event])
+    [:pupil, :coach, :organizer].each do |role|
```
frederike-ramin, line 40: Coaches and organizers are not supposed to be able to apply at all, right? #286

app/views/events/_apply_button.erb
```diff
@@ -0,0 +1,7 @@
...
+                :class => 'btn btn-default btn-sm' %>
+<% elsif not event.after_deadline? %>
```
frederike-ramin, line 4: Please add check of :apply ability once #286 is in dev.

frederike-ramin (changes requested):
Maybe you should talk to Lando because of the deny application for organizers issue.

app/models/user.rb
querenker: I am not sure. [This style guide](https://github.com/airbnb/ruby#no-and-or) for example is strictly against the usage of not. The reason is that the `not` operation has a very low precedence whereas the `!` has one of the highest precedence just like I would expect from other languages.

app/views/events/_apply_button.erb
querenker: From the [user story description](https://github.com/hpi-swt2/workshop-portal/issues/244):

> When a user applied for an event the "Bewerben" buttons should be replace with "Bewerbung anzeigen" and link to the application checking page.

app/views/events/_apply_button.erb
querenker: Thanks for the hint. But is it really necessary? I think if a person can not apply there should not even be an option to create an application letter for them?

spec/views/events/index.html.erb_spec.rb
querenker: This test should ensure that the apply button is hidden if the deadline is over (independent of the role of the user). But you are right if there exists test that guarantee that coaches and organizers can not apply that is a little bit redundant.

spec/features/events_spec.rb
querenker: The problem is that I need a event that is in selection phase and has exactly one application. If a event in selection phase is created no application letter can be added because the deadline is over...

spec/views/events/index.html.erb_spec.rb
bjrne: I would still leave it like it is, since somewhere in the future it might be necessary to let roles different from pupil apply to events too. I guess removing roles because they shouldn't be able to apply would mix logic for two things in a bad way.

app/views/events/_apply_button.erb
frederike-ramin: Yes, but it's not really consistent to have a button that doesn't work?!

spec/features/events_spec.rb
frederike-ramin: Although you COULD add a factory if you'd want to.

app/views/events/_apply_button.erb
querenker: This `current_user.application_letter_for_event?(event)` checks if for the current user exists an application letter for the event. If no application letter exists the button will not be displayed.

spec/features/events_spec.rb
querenker: @frederike-ramin Done

frederike-ramin (approved):

Issue #283, opened 2017-01-10
US_2.43: Application Delete ButDann wurde der Grundstein tons


hendraet:
**As**
admin

**I want**
to be able to delete applications of users but no one should be able to edit applications except the user itself

**In order to**
clean up mistakes or remove spam/unecessary applications

**Dependencies:**

**Estimate:** 5

**Acceptance Criteria:**
- [x] delete button on the details page for an application should only be visible to admins and no one else
- [x] edit button should also be remove bc. organiser shouldn't have access to this page 
- [x] should be displayed at the bottom
- [x] should look like this:
![image](https://cloud.githubusercontent.com/assets/16731894/21811714/fa6c9ce6-d750-11e6-9ab5-2546f871dbbc.png)

PR #390, opened 2017-01-23
#283 2.43 application delete buttons

hefowe:
ref #283


spec/views/application_letters/show.html.erb_spec.rb
```diff
@@ -1,10 +1,14 @@
...
+require 'request_helper'
 
 RSpec.describe "application_letters/show", type: :view do
   before(:each) do
     @application_letter = assign(:application_letter, FactoryGirl.create(:application_letter))
     @application_note = assign(:application_note, FactoryGirl.create(:application_note, application_letter: @application_letter))
     @application_letter.user.profile = FactoryGirl.build(:profile)
+
+    user = FactoryGirl.create(:user, role: :admin)
+    sign_in user
```
cmfcmf: I would prefer to login as `@application_letter.user`.

app/views/application_letters/show.html.erb
```diff
@@ -83,3 +81,11 @@
       </div>
     <% end %>
 <% end %>
+
+<% if current_user.role == "admin" %>
```
cmfcmf: Why can't you use `can? :delete, ApplicationLetter`?

cmfcmf (changes requested):
LGTM, just need some clarifications.

cmfcmf (approved):
Changes look good! 🎉 👍

PR #397, opened 2017-01-24
Add env vars to configure production host und port. Fixes #299.

### LINKED ISSUE #299: Forgot Password functionality does not work

tom95:
Last branch relied on a static field on ActionMailer::Base which seems to have been removed in a more recent version of rails, this one doesn't depend on anything. Tested by copying over to development environment.


bjrne (approved):
Looking forward to it working now 🐝

Issue #339, opened 2017-01-17
US_3.29: Show email address in profile


tobiasduerschmid:
**As**
organizer and admin

**I want to**
see the email address of the user in his / her profile

**in order to**
write an email to him / her

# Tasks

# Acceptance criteria

- When an admin or an organizer visits the profile page of a user, the email address is displayed. 

- The text is hyperlinked to a mailto: link

PR #315, opened 2017-01-11
125 3.20 hide information

Baschdl:
This PR hides non-application-related information on the application letter page (for coaches and organizers) and displays a link to the profile of an applicant (for organisers and admins). Fixes #125

PR #398, opened 2017-01-24
[WIP] 339 3.29 email address in profile

Baschdl:
The email address is displayed on the profile page. Only organizers and admins can view the profile page of applicants so we don't need an extra ability. Refers to #339


ekrebs5:
Why did you write feature tests and not view tests?

From what I see you added something to the view and because of separate ability stuff only the right people see it, right?
So I would have expected a view test with "email is displayed" (and maybe ability/feature tests for who can access the profile, but those should already be implemented if the ability already exists?)

Baschdl:
@ekrebs5 @cmfcmf Yeah, that's right, but as in #315 I was unable to login the user with the right role in a view test. Do you know a solution for this problem?

ekrebs5:
in the view tests in my pr I use something like
```
@event = assign(:event, FactoryGirl.create(:event, :in_selection_phase))
sign_in(FactoryGirl.create(:user, role: :organizer))
```

That said, since you only added to the view, should't something like the existing "renders attributes" test work?

Baschdl:
Ok, I will try that

spec/features/profile_spec.rb
```diff
@@ -95,6 +95,38 @@ def mock_writing_to_filesystem
...
 end
 
+RSpec.feature "Visit page with special role", :type => :feature do
+  before :each do
+    @profile = FactoryGirl.create(:profile)
+    @user = FactoryGirl.create(:user, role: :pupil, profile: @profile)
+    @event = FactoryGirl.create(:event)
+    @application_letter = FactoryGirl.create(:application_letter_accepted, user: @user, event: @event)
+    login_as(@user, scope: :user)
+    visit profile_path(@user.profile)
```
aloeser, line 105: I think you can drop this line and the three lines above.
You dont use events/application letters here, and since you are testing with different roles you need to re-login and visit each at the beginning of each test anyway, so we dont need it here

aloeser: you might be able to skip the user too, if you just need the profile. Not really sure, but since profile belongs to a user, one should be auto-created if you create a profile (maybe that changed tho, you gonna have to try it out)

aloeser (changes requested):
As @ekrebs5 said, you are missing view tests.
Also, what does the [WIP] stand for (what is missing)?

In the PR description, you say "Only organizers and admins can view the profile page" -  what is with the participants themselves?

frederike-ramin (changes requested):
I don't really see how this is WIP?
But I agree with @aloeser about the unnecessary lines in the tests.

spec/features/profile_spec.rb
Baschdl: Yes, too much copy&paste

Baschdl:
Sort of WIP, I must convert the feature tests into view tests

Baschdl:
I'll close it for now

Issue #341, opened 2017-01-17
US_2.4.6: Hidden Events


hendraet:
**As**
organiser

**I want**
to be able to create events which aren't publicly visible but to which one can apply via link

**In order to**
be able to apply to requested workshops

**Dependencies:**

**Estimate:** 5

**Acceptance Criteria:**
- [x] additional checkbox on workshop creation page: "versteckt" (under Workshop/camp  #Radio-Button)
- [x] button should be right next to the other buttons at the top
- [x] event should only be visible to organiser on event page
- [x] there should still be a check, if user has a profile and is logged in (if possible)
- [ ] link should point directly to application page

PR #400, opened 2017-01-24
#341 2.4.6 hidden events

nilsthamm:
ref #341

### DIFF
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -8,12 +8,8 @@ class EventsController < ApplicationController
   # GET /events
   def index
     @events = Event.sorted_by_start_date(false)
-    if !can? :view_hidden, Event
-      @events = @events.select{|event| event.hidden == false}
-    end
-    if !can? :view_unpublished, Event
-      @events = @events.select{|event| event.published == true}
-    end
+    @events = @events.select{|event| event.hidden == false} unless can? :view_hidden, Event
+    @events = @events.select{|event| event.published == true} unless can? :view_unpublished, Event
   end
 
   # GET /events/1
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -7,11 +7,20 @@ class EventsController < ApplicationController
 
   # GET /events
   def index
-    @events = Event.sorted_by_start_date(!can?(:view_unpublished, Event))
+    @events = Event.sorted_by_start_date(false)
+    if !can? :view_hidden, Event
+      @events = @events.select{|event| event.hidden == false}
+    end
+    if !can? :view_unpublished, Event
+      @events = @events.select{|event| event.published == true}
+    end
   end
 
   # GET /events/1
   def show
+    if @event.hidden and !can? :view_hidden, Event
+      redirect_to new_application_letter_path(:event_id => @event.id)
+    end
     @free_places = @event.compute_free_places
     @occupied_places = @event.compute_occupied_places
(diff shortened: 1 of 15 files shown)


app/controllers/events_controller.rb
```diff
@@ -7,11 +7,20 @@ class EventsController < ApplicationController
 
   # GET /events
   def index
-    @events = Event.sorted_by_start_date(!can?(:view_unpublished, Event))
+    @events = Event.sorted_by_start_date(false)
+    if !can? :view_hidden, Event
+      @events = @events.select{|event| event.hidden == false}
+    end
+    if !can? :view_unpublished, Event
```
cmfcmf: This is unrelated to the PR, right? I assume this was a bug?

app/controllers/events_controller.rb
```diff
@@ -7,11 +7,20 @@ class EventsController < ApplicationController
 
   # GET /events
   def index
-    @events = Event.sorted_by_start_date(!can?(:view_unpublished, Event))
+    @events = Event.sorted_by_start_date(false)
+    if !can? :view_hidden, Event
```
cmfcmf: nit: You could use `unless`.

app/controllers/events_controller.rb
cmfcmf: nit: This could also benefit from `unless`.

app/views/events/_form.html.erb
```diff
@@ -13,7 +13,13 @@
       <%=f.error_span(:kind) %>
     </div>
   </div>
-
+  <div class="form-group">
+    <div class="col-lg-offset-2 col-lg-10">
+      <%= f.check_box :hidden%>
```
cmfcmf: nit: missing space before `%>`

app/views/events/_form.html.erb
```diff
@@ -13,7 +13,13 @@
...
+      <%= f.label :hidden, :class => 'control-label' %>
+      <%=f.error_span(:hidden) %>
```
cmfcmf: nit: missing space after `<%=`

spec/controllers/events_controller_spec.rb
```diff
@@ -165,6 +165,17 @@
     end
   end
 
+  describe "GET #show for hidden event as pupil"
+    it "should redirect to new application letter page" do
+      @event = FactoryGirl.create(:event, hidden: true)
+      @user = FactoryGirl.create(:user, role: :pupil)
+      @user.profile ||= FactoryGirl.create(:profile)
```
cmfcmf: Use `:user_with_profile` instead

app/controllers/events_controller.rb
cmfcmf: > This is unrelated to the PR, right? I assume this was a bug?

Also, I don't see a new test for this feature.

cmfcmf (changes requested):
Mostly minor nitpicks, looks good overall!

app/controllers/events_controller.rb
nilsthamm: > This is unrelated to the PR, right? I assume this was a bug?

the unpuplished events were filtered before by the sorted_by_start_date function (so this is already tested in spec/features/event_spec.rb:28). 
Because i wanted to distinguish between hidden and unpuplished events, I test the rigths to view them separately. I know at the moment, this distinction makes no difference, but i don't think it is a good idea to summerize it to one ability.

cmfcmf (approved):
LGTM now!

Issue #242, opened 2017-01-05
US_3.25: Application canceled state


tobiduer:
**As**
organizer

**I want to**
change participants application state to canceled

**in order to**
be able to choose an alternative applicant who was rejected. 

# Tasks

## New state:
- [x] new canceled + pre_accepted state
- [x] refactor code accordingly
- [x] add pre_accepted application_letter factory
- [x] rewrite (accepted/rejected)_applications_count

- [x] rename Event accept_all_application_letters
- [x] update accepted_application_count to use both accepted states
- [x] rename application_letter_applicant_gongakrobatic_accepted
- [x] make sure, that you can filter after all statuses (pretty much refactor everything in that test) features/events_spec filter displayed application_letters
- [x] has as many participants as accepted applications
- [x] click on accept radio button --> pre_accept
- [x] checks states of applications
- [x] coaches see pre_accepted status
- [x] make only accepted, rejected, pending and alternative selectable, not cancled and pre_accepted
- [ ] pre_accepted -> accepted
- [x] rename draft status to acceptance and rejection
- [ ] test states
- [x] selectable_statuses as Application_letter variable

## Actually do stuff:
- [x] cancel button for accepted applications
- [x] cancel when clicked
- [x] make accept button appear (alternatives)
- [x] only accept button when counter good (alternatives)
- [x] pre_accept, when clicked (alternatives)
- [ ] make accept button appear (rejected)
- [ ] only accept button when counter good (rejected)
- [ ] pre_accept, when clicked (rejected)
- [ ] make email button appear (pre_accepted)
- [x] ASK PO TO SEND ALL EMAILS INSTEAD OF LETTER ICON
- [x] ASK PO IF ONLY RECIPIENTS OF EMAIL SHOULD FINALLY BE ACCEPTED
- [ ] redirect to mail sending acceptance email when clicked
- [ ] accept ONLY recipients of email after mail button clicked
# Acceptance criteria

- When the event is in state execution, organizers should see a button "cancel" in the row of each accepted participant. 

- When the organizer pushed the button, then the application is canceled, meaning that the applicant will not participate in the event (similar to rejected) 

- After pushing the cancel button, the organizer can choose a new participant out of the list of alternatives (Nachrücker) and if no alternatives present, then a list of rejected applicants. 

- Applicants who have been accepted in the execution state but did not get an email yet have an emal icon next to the "angenommen" label that lead to the page for sending the acceptance email to him / her. 

[...]

PR #401, opened 2017-01-24
Rename Email Template statuses #242

aloeser:
This PR introduces better names for the email template status. Until now, the application letter statuses where used, now we have a better distinction between application letter status and template status


app/controllers/emails_controller.rb
```diff
@@ -54,15 +54,22 @@ def save_template
...
-  def get_status
+  def get_email_template_status
     params[:status] ? params[:status].to_sym : :default
   end
 
+  def get_corresponding_application_letter_status
+    return :accepted if params[:status] == "acceptance"
+    return :rejected if params[:status] == "rejection"
+    # default value
+    return :accepted
```
querenker: I think a case statement would be slightly better to read (and you get rid of the comment).

app/controllers/emails_controller.rb
```diff
@@ -54,15 +54,22 @@ def save_template
...
 
     render :email
   end
 
-  def get_status
+  def get_email_template_status
     params[:status] ? params[:status].to_sym : :default
   end
 
+  def get_corresponding_application_letter_status
```
querenker, line 66: YARD annotation is missing

app/controllers/emails_controller.rb
```diff
@@ -54,15 +54,22 @@ def save_template
...
     end
     @event = Event.find(params[:event_id])
-    @templates = EmailTemplate.with_status(get_status)
+    @templates = EmailTemplate.with_status(get_email_template_status)
 
     render :email
   end
 
-  def get_status
+  def get_email_template_status
```
querenker, line 62: YARD annotation is missing

app/controllers/emails_controller.rb
querenker, line 66: Currently maybe it looks a little bit overkill but a unit test should be provided, shouldn't it?

querenker (changes requested):
Otherwise it looks good for me :)

app/controllers/emails_controller.rb
aloeser: Do you mean the documentation stuff? I think we agreed to write documentation for model functions, and this is just a (private) helper function inside the controller

app/controllers/emails_controller.rb
aloeser: same

aloeser: Usually yes, but its a private controller helper. Calling this function alone doesnt make sense at all (and its not ment to be accessible from "outside". And I think that we never tested private helper methods (like e.g. event_params or such), so it seems inconsistent for me if we add a test for this. Instead, we test the public function that calls it.

app/controllers/emails_controller.rb
querenker: You are right, a private method doesn't need documentation stuff

app/controllers/emails_controller.rb
querenker: You are right, we rarely test private methods.  But I think test should beside guaranteing that the public interface works, minimize the effort to find the bug if something got broken (I don't like debugging). If the private method would break a test it would point directly to the broken location. I think for consistency reasons it is OK in this PR, but maybe we could talk about it later? @Baschdl?

querenker (approved):

app/controllers/emails_controller.rb
```diff
@@ -54,15 +54,26 @@ def save_template
...
+  def get_corresponding_application_letter_status
+    case params[:status]
+    when "acceptance"
+      return :accepted
+    when "rejection"
+      return :rejected
+    else
+      return :accepted
+    end
+  end
```
querenker, line 72: Just as little improvement. I think the case statement could also be written that way:
```ruby
return case params[:status]
       when "acceptance" then :accepted
       when "rejection" then :rejected
       else :accepted
       end
```

aloeser: we can change it like that, looks easier to read

ekrebs5 (approved):

PR #405, opened 2017-01-24
Make sure send mails buttons are updated via xhr. Fixes #377.

### LINKED ISSUE #377: Clicking on radio buttons should reload email buttons
Since #128 clicking on radio buttons does not reload the entire page.
This includes the buttons, which can lead to the email buttons being in a wrong state.

Example:
I have a workshop with a 42 participant maximum.
When I go to the page, 42 people are marked as accepted and 1 as rejected. The email buttons are clickable, like they should be.
Now I change the 1 person also to accepted. The buttons stay clickable (and lead to the mail sending page) even though they should not (since I now have 43 people for 42 people workshop).

In short: The email buttons should also be reloaded when I click on the radio buttons

tom95:


cmfcmf (changes requested):
Wach out: disabled does not work properly for links. Also, the template was heavily changed on dev already.

cmfcmf (approved):

Issue #46, opened 2016-11-18
US_1.7: Additional Information 🍪7


Helene0:
**As** 
organizer 

**I want to** 
be able to add the name of the external organisation, if there is one (Veranstalter), and the knowledge pupils need to participate (Kenntnisstand), but only if needed

**in order to** 
choose what information is relevant for pupils

**7 🍪**

- [x] The fields "Veranstalter" and "Kenntnisstand" are optional and also appear as such.
- [x] I can enter text in those fields and save it

MasterCarl:
See merged pull request https://github.com/hpi-swt2/workshop-portal/pull/71

MasterCarl:
Also https://github.com/hpi-swt2/workshop-portal/pull/80

PR #412, opened 2017-01-25
238 event description textfield

corinnaj:
Adjusted Multiline Placeholder so that it also works on other browsers, such as Firefox.


app/assets/javascripts/events.js
```diff
@@ -3,14 +3,28 @@
...
+    var list = button.data('list');
+    var modal = $(this);
+    modal.find('.modal-title').text(header);
+    modal.find('#send-emails-mailto').attr('href', 'mailto:' + list);
+    modal.find('#send-emails-list').val(list);
+  });
+
+  // work around so that we can have a multiline placeholder
+  $("#description")
+    .val(function() { return PLACEHOLDER; }).css('color', '#bbb')
```
tom95: This part would override existing content in the text field. You could use
```
.val(function(index, oldValue) { return oldValue || PLACEHOLDER; });
```
to fix this.

Additionally, jQuery() onload won't trigger for turbolinks navigation. You could try if turbolinks-jquery will fix that or alternatively use the same logic as found in `app/assets/javascripts/bootstrap.js` to make sure the code is triggered on turbolink loads as well.

app/views/events/_form.html.erb
```diff
@@ -31,22 +31,9 @@
...
-                $(this).attr('placeholder', '');
-            }
-        });
-
-        $("#description").blur(function(){
-            if($(this).val() ===''){
-                $(this).attr('placeholder', placeholder);
-            }    
-        });
+        var PLACEHOLDER = '<%= I18n.t "events.description_placeholder" %>';
```
tom95: Maybe prefix the constant name as `EVENT_DESCRIPTION_PLACEHOLDER`, just in case anyone else was to introduce another placeholder somewhere.

tom95 (changes requested):

MasterCarl (approved):
It works now, looks good.

cmfcmf (commented):
I can't review this at the moment, because I'm getting the following error when trying to create a new event: 
> Showing /home/vagrant/hpi-swt2/app/views/events/_form.html.erb where line #46 raised:
> undefined local variable or method `participants_are_unlimited' for #<Event:0x007fa44fb90b90>
```
 def max_participants
    participants_are_unlimited ? Float::INFINITY : self[:max_participants]
  end
```

bjrne:
I guess this needs a new merge from dev, but I'm not sure I'm qualified to fix the conflicts. Could you give this a look @corinnaj @tom95?

tom95 (approved):

PR #414, opened 2017-01-25
Add archive page for past events. Fixes #304.

### LINKED ISSUE #304: Archive Page For Past Events

tom95:


spec/factories/date_ranges.rb
```diff
@@ -28,5 +28,11 @@
             start_date Date.tomorrow
             end_date Date.tomorrow
         end
+
+        trait :force_past do
```
bjrne: I'm not completely sure about RoR naming conditions, but I think "Entity :withX :withoutY" etc reads better than these titles.

app/views/events/index.html.erb
```diff
@@ -6,14 +6,18 @@
 
 <div class="row">
   <div class="event-list col-md-10 col-md-offset-1">
-    <% if can? :new, Event %>
     <div class="clearfix">
       <%= link_to new_event_path, :class => 'btn btn-primary pull-right' do
         glyph('plus') + ' ' +
         t('.new', :default => t('helpers.links.new'))
+      end if can? :new, Event %>
```
MasterCarl, line 13: Please add a newline between `end` and `if`

app/views/events/archive.html.erb
```diff
@@ -0,0 +1,12 @@
+<div class="row">
+  <div class="event-list col-md-10 col-md-offset-1">
+    <%= link_to events_path, :class => 'btn btn-default' do
+      glyph('chevron-left') + ' ' +
+      t('.new', :default => t('events.all_events'))
```
MasterCarl: Why is the translation key `.new`?

app/controllers/events_controller.rb
```diff
@@ -254,6 +256,13 @@ def event_params
 
     end
 
+    def add_event_query_conditions(query)
```
MasterCarl, line 259: I'm not too happy with the naming here, but then again, I don't have any better ideas, either.

MasterCarl (changes requested):
Overall, this looks good

app/views/events/index.html.erb
tom95: this is actually a "suffix" if. believe me if I tell you that it was even more confusing the other way around

app/controllers/events_controller.rb
corinnaj: I think the naming is fine.

app/views/events/index.html.erb
MasterCarl: Oh, alright

MasterCarl (approved):

PR #419, opened 2017-01-25
QUICKFIX make email button visible in execution phase

ekrebs5:
Quickfix for presentation on friday

Emails buttons are now also visible in execution phase, so you can always send both mails

This should be changed (execution phase starts after both mails are sent) in the future

### DIFF
--- a/app/views/events/_applicants_overview.html.erb
+++ b/app/views/events/_applicants_overview.html.erb
@@ -66,7 +66,7 @@
     </div>
 
     <div class="btn-group pull-right" role="group">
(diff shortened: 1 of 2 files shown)


bjrne:
I don't really like the idea to add logic just for presentation purposes. Is this really necessary?

ekrebs5:
@tobiduer asked for this
at our metting yesterday we noticed that with the current state system it would only be possible to send one kind of mail (acceptions or rejections)

ekrebs5:
Ok, @querenker talked with our po @tobiduer today and he does not want it anymore

Issue #179, opened 2016-12-11
US_4.13: "Wie hast du von uns erfahren" Field in Create Profile 💰 5


hendraet:
**As**
organiser

**I want**
pupils to enter the "Wie hast du von erfahren" field during create profile

**In order to**
get more data

**Acceptance Criteria:**
- [ ] delete field from application page
- [x] new field in registration
- [x] field is not mandatory
- [x] data is saved in database
- [x] can be modified in profile
- [x] show for organizer the information in the user profile

**Tasks**
- [x] add column to db
- [x] modify form, controller, model
- [x] show for organizer the information in the user profile

julisa99:
forgot to delete the field in application form

Fabian-Sommer:
Why can't this be accepted @tobiduer?

PR #435, opened 2017-01-28
179 4.13 Added discovery_of_site field to profile

Fabian-Sommer:
#179

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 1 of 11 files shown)


db/schema.rb
```diff
@@ -112,6 +112,7 @@
     t.string   "state"
     t.string   "country"
     t.string   "graduates_school_in"
+    t.text     "discovery_of_site"
```
pinussilvestrus, line 115: is there a migration missing? when I run `db:migrate`, it's missing

pinussilvestrus: image

pinussilvestrus (approved):

PR #436, opened 2017-01-28
Add text to index page, include translations

MasterCarl:
References #350


app/views/application/index.html.erb
```diff
@@ -130,3 +130,25 @@
...
+				<div class="row">
+					<div class="col-xs-12 col-sm-12 col-md-12 col-lg-12 text-center">
+						<h2><%= t('start_page.about_hpi.title') %></h2>
+					</div>
+				</div>
+				<div class="row">
+					<div class="col-xs-12 col-sm-12 col-md-12 col-lg-12">
+						<p class="lead">
+							<%= t('start_page.about_hpi.lead_before_link') %>
+							<a href="hpi.de"><%= t('start_page.about_hpi.lead_with_link') %></a>
```
bjrne: This leads to the following error when clicked on locally: 

> `No route matches [GET] "/hpi.de"`

I guess you need to enter the full path. It works with `https://www.hpi.de` and `https://hpi.de` though. Is there a reason for one or the other?

tom95 (approved):

Issue #392, opened 2017-01-23
Complete the "Anfrage"-detail page


julisa99:
**As**
organizer

**I want to**
see all information of a request

**In order to**
be able to correspond with the requesting person and get the details about the request. 

## Acceptance criteria:
- When a Alice created a request, organizers see all information, Alice entered in the form "Anfrage-stellen"-Page on the "Anfrage"-detail page. Currently for example missing: E-Mail, Orga/Firma, ...
- The email address is linked with a Mail-to-Link in order to simplify writing an E-Mail.

julisa99:
@MasterCarl Zeitrahmen wird nicht angezeigt auf der detail-page

MasterCarl:
Fixed in #472

PR #437, opened 2017-01-28
Complete the "Anfrage"-detail page

MasterCarl:
References #392


app/views/requests/show.html.erb
```diff
@@ -1,19 +1,20 @@
 <%- model_class = Request -%>
 <div class="page-header">
-  <h1><%=t '.title', :default => model_class.model_name.human(count: 2).titleize %></h1>
+  <h1><%=t '.title', :default => model_class.model_name.human(count: 1).titleize %></h1>
```
bjrne: Do you need `count` if you only want the singular?

bjrne (approved):
Rest looks good 🐎

Issue #319, opened 2017-01-12
Workshop Choice Bar


Helene0:
**As**
user

**I want to**
have a slidebar with example workshop pictures and Topics for inspiration

**In order to**
know what kind of ws i could ask for

**Acceptance Criteria:**

- [ ] I have a display of options, ideally with pictures and a Title (topic) 
- [ ] OPTIONAL this could be in form of a slidebar

radscheit:
@hendraet Please have a look on the screener attached. Is the first orange panel fitting to your expectations?

![bildschirmfoto 2017-01-28 um 00 47 13](https://cloud.githubusercontent.com/assets/6578648/22397313/0c6b150e-e56f-11e6-9858-4401146174e0.png)

hendraet:
@radscheit Sorry, but I don't see an attachement

radscheit:
@hendraet This is already a PR #438. Please check it, since the acceptance criteria are letting room for interpretation.

julisa99:
@hendraet ?

hendraet:
@julisa99 @radscheit Told Matthias that he should merge it into dev so I can review it.

radscheit:
@julisa99 @hendraet Sorry I was distracted by my birthday and my final presentation for Dschool that will take place today. I will push it this night. :)

radscheit:
@julisa99 @hendraet I just noticed that i already pushed it days ago: https://github.com/hpi-swt2/workshop-portal/pull/438 :)

julisa99:
@radscheit really? I don't see it
<img width="729" alt="anfrage" src="https://cloud.githubusercontent.com/assets/8758907/22648984/4b29e0a2-ec79-11e6-9304-195ae6c90ce2.PNG">

radscheit:
@julisa99 Wo hast du es dir denn angeschaut?

hendraet:
@radscheit Sorry, but I have to reopen this issue. The autoscrolling is a nice idea but is just annoying when you try to read the text. Please remove it.

hendraet:
@radscheit an please exchange the text. Just copy them from here: https://hpi.de/open-campus/schuelerakademie/projekttage.html

radscheit:
@hendraet As an alternative for dropping autoplay, i could offer stopAutoplay on mousehover?

hendraet:
Passt

PR #438, opened 2017-01-28
319 workshop choice bar

radscheit:
(#319)


app/assets/stylesheets/requests.css
```diff
@@ -2,3 +2,62 @@
   Place all the styles related to the matching controller here.
   They will automatically be included in application.css.
 */
+
+.owl-theme .owl-controls{
+margin-top:10px;
```
pinussilvestrus: can you add tabs in this stylesheet?

config/locales/de.requests.yml
```diff
@@ -2,6 +2,43 @@
 
 de:
   requests:
+    templates:
+      headline: "Beispiel-Anfragen"
+      lead_paragraph: "Lorem ipsum dolor sit amet, consectetur adipisicing elit. Laborum fugiat quasi expedita distinctio, suscipit, eaque deleniti similique eveniet in accusantium voluptas blanditiis id modi voluptatibus culpa illum aliquam. Neque, dolore."
```
pinussilvestrus, line 7: I don't know whether there is another story for it, but wouldn't it be better if here is a better content than Lorem ipsum?

pinussilvestrus (changes requested):
image

this doesn't look correct or?

config/locales/de.requests.yml
radscheit: Done.

pinussilvestrus (approved):

Issue #271, opened 2017-01-09
US_4.14: Additional field in application form 💰 5


julisa99:
**As**
pupil

**I want to**
add additional information or questions

**In order to**
have the opportunity to do it there.

**Acceptance criteria**
- [x] additional field at the end of the application form
- [x] name of the field is "Hast du weitere Anmerkungen oder Fragen?"
- [x] show it in printed pdf

Issue #374, opened 2017-01-21
Event feature test fails randomly


Blaidd-Drwg:
'Event application letters overview on event page logged in as Organizer I can filter displayed application letters by their status and simultaneously sort them' sometimes fails with the message 'expected #<Capybara::Session> to contain ordered "Karl100 Doe100", "Karl101 Doe101", "Karl102 Doe102", "Karl98 Doe98", and "Karl99 Doe99"' at [this line](https://github.com/hpi-swt2/workshop-portal/blob/5e12a7d041a4019d84bd2d75b3b043fac7931716/spec/features/events_spec.rb#L299).

From what I've gathered so far, the names are sorted by id in the HTML (...Karl99, Karl100...), yet the test expects them to be sorted alphabetically (...Karl100, Karl99...).

Blaidd-Drwg:
[`contain_ordered`](https://github.com/hpi-swt2/workshop-portal/blob/5e12a7d041a4019d84bd2d75b3b043fac7931716/spec/support/matchers/ordered_appearance.rb) does not fail consistently, sometimes (seemingly at random) it passes even though it shouldn't. @MasterCarl Any idea why?

MasterCarl:
That's odd, I've never had these tests fail. Honestly, I have no idea. Does this only happen locally?

Blaidd-Drwg:
No, on Travis as well. See #371: continuous-integration/travis-ci/push and continuous-integration/travis-ci/pr both failed at first. I restarted continuous-integration/travis-ci/pr without any changes to the branch and suddenly it passed.

As far as I can see, the problems are two:
1) In the test, names are **never** sorted alphabetically, which should cause it to fail every time.
2) In same cases, though, `contain_ordered` still passes, which is why our build status on dev is currently green.

tom95:
Could you provide a link to the travis build in question? I had this happen too, but I thought I fixed it on dev.

Blaidd-Drwg:
The one that failed: https://travis-ci.org/hpi-swt2/workshop-portal/builds/193932385
The one that succeeded later on: https://travis-ci.org/hpi-swt2/workshop-portal/builds/193932389

I haven't merged your commits into that branch, so the builds don't include your fix. But since I'm having the exact same problem when testing the current dev locally, I figure it's still an issue on Travis as well.

MasterCarl:
[...]

PR #440, opened 2017-01-29
271 4.14 additional application field

phnm:
#271


app/controllers/emails_controller.rb
```diff
@@ -7,7 +7,7 @@ def show
     @templates = EmailTemplate.with_status(get_status)
     @addresses = @event.email_addresses_of_type(get_status)
 
-    @email = Email.new(hide_recipients: true, reply_to: 'workshop.portal@hpi.de', recipients: @addresses,
+    @email = Email.new(hide_recipients: true, reply_to: 'workshop.portal@hpi.de', recipients: @addresses.join(','),
```
bjrne, line 10: How is this change related to the topic of this PR?

db/migrate/20170124225558_add_annotation_to_application_letters.rb
```diff
@@ -0,0 +1,5 @@
+class AddAnnotationToApplicationLetters < ActiveRecord::Migration
+  def change
+    add_column :application_letters, :annotation, :string
```
bjrne: Please use text here, this could end in a lot more than 255 characters

spec/controllers/events_controller_spec.rb
```diff
@@ -217,33 +217,33 @@
...
+
       user = FactoryGirl.create(:user)
       profile = FactoryGirl.create(:profile, user: user, last_name: "Peter")
-      application_letter = FactoryGirl.create(:application_letter_accepted, 
+      application_letter = FactoryGirl.create(:application_letter_accepted,
         user: user, event: event, vegan: true)
       user = FactoryGirl.create(:user)
       profile = FactoryGirl.create(:profile, user: user, last_name: "Paul")
-      application_letter = FactoryGirl.create(:application_letter_accepted, 
+      application_letter = FactoryGirl.create(:application_letter_accepted,
```
bjrne, line 227: Please don't commit so many whitespace changes

spec/features/participants_spec.rb
```diff
@@ -47,48 +47,40 @@
...
       user: user, event: @event, vegetarian: true)
     user = FactoryGirl.create(:user)
     profile = FactoryGirl.create(:profile, user: user, last_name: "Otti")
-    application_letter = FactoryGirl.create(:application_letter_accepted, 
+    application_letter = FactoryGirl.create(:application_letter_accepted,
       user: user, event: @event, vegetarian: true, allergic: true)
     user = FactoryGirl.create(:user)
     profile = FactoryGirl.create(:profile, user: user, last_name: "Benno")
-    application_letter = FactoryGirl.create(:application_letter_accepted, 
+    application_letter = FactoryGirl.create(:application_letter_accepted,
```
bjrne: same here

spec/models/event_spec.rb
```diff
@@ -91,8 +91,8 @@
     accepted_application_letter_3 = FactoryGirl.create(:application_letter_accepted, :event => event, :user => FactoryGirl.create(:user))
     rejected_application_letter = FactoryGirl.create(:application_letter_rejected, :event => event, :user => FactoryGirl.create(:user))
     [accepted_application_letter_1, accepted_application_letter_2, accepted_application_letter_3, rejected_application_letter].each { |letter| event.application_letters.push(letter) }
-    expect(event.email_addresses_of_type(:accepted)).to eq([accepted_application_letter_1.user.email, accepted_application_letter_2.user.email, accepted_application_letter_3.user.email].join(','))
-    expect(event.email_addresses_of_type(:rejected)).to eq([rejected_application_letter.user.email].join(','))
+    expect(event.email_addresses_of_type(:accepted)).to contain_exactly(accepted_application_letter_1.user.email, accepted_application_letter_2.user.email, accepted_application_letter_3.user.email)
+    expect(event.email_addresses_of_type(:rejected)).to contain_exactly(rejected_application_letter.user.email)
```
bjrne, line 95: And why is this change in this PR. I don't see it related to this in any way

spec/models/event_spec.rb
```diff
@@ -186,21 +186,16 @@
...
     end
-    
+
     it "returns email address only of the given type" do
       @accepted_application = FactoryGirl.create(:application_letter_accepted, event: @event, user: FactoryGirl.create(:user))
       @rejected_application = FactoryGirl.create(:application_letter_rejected, event: @event, user: FactoryGirl.create(:user))
-      expect(@event.email_addresses_of_type(:accepted)).to eq(@accepted_application.user.email)
-      expect(@event.email_addresses_of_type(:rejected)).to eq(@rejected_application.user.email)
+      expect(@event.email_addresses_of_type(:accepted)).to contain_exactly(@accepted_application.user.email)
+      expect(@event.email_addresses_of_type(:rejected)).to contain_exactly(@rejected_application.user.email)
```
bjrne, line 194: again. Why in this PR? Please open separate PRs for separate topics.

bjrne (changes requested):
I would approve this, if only features related to your issue were changed. However I don't like all the whitespace changes and things that have nothing to do with this. Or is this just a mistake and those changes came from dev?

spec/models/event_spec.rb
phnm: You're right. But Travis does. This is a long story, but for some reason 3 tests were always failing only on travis and always passing on local. Even though I didn't touch the test or the feature they test, as you can see in the commit history of my branch. Don't ask me why, but I had to refactor those tests in order to have them passing on Travis.

spec/controllers/events_controller_spec.rb
phnm: My editor (Atom) automatically deletes trailing whitespaces in files I touch. Is that a bad thing?

db/migrate/20170124225558_add_annotation_to_application_letters.rb
phnm: Good point, I will change this.

app/controllers/emails_controller.rb
phnm: I had to refactor the email_addresses_of_type(type) method, since it was returning a string with concatenated emails addresses in a random order on Travis. Only Travis knows why this wasn't an issue earlier, but tests were suddenly faling because of this. I changed it to return an array of Emails, which is better anyway, but made this change to join the addresses here necessary.

spec/models/event_spec.rb
bjrne: I'd have liked to see a separate small PR for this, followed by the feature one. This way, the 'travis'-PR would have maybe attracted some other people for reviewing and commenting (and noticing) than hiding it in this PR did. Things like this may be of interest for more developer. Do you know now exactly why this change solves the problem?

spec/controllers/events_controller_spec.rb
bjrne: not necessarily, but I like to only have the productive changes in the feature PRs in order to be able to focus on those. If people constantly leave trailing whitespace and editors change it forth and back it might be a good idea to have a cleanup PR where you fix a lot of those formatting problems.

app/controllers/emails_controller.rb
bjrne: Hm, this is more refactoring than I'd like in a PR for another topic. I'd like some feedback from somebody else, maybe from your team or from the email people @frederike-ramin @aloeser @Baschdl. I would really like this to be separated from this PR. It shouldn't be too much effort with `git stash`.

spec/models/event_spec.rb
phnm: I know what went wrong, but have no idea why now and why only on travis. I also posted this in issue #374 since this might be related, but nobody seemed to have an idea.
The problem with this test was that, when comparing a string with `eq` there is an assumed order of the expected concatenated email addresses. But those addresses were never sorted, they just happend to be in the expected order so far. And that order was different on Travis on my branch.

bjrne: Hm, okay. Would you be willing to open a separate PR for this, link it to the issue and once it is merged, we merge this one? I feel like these are two full features, since your refactoring also changed a little bit of logic.

phnm: Sure, see #441

bjrne (approved):
Looks good now 🏎

PR #441, opened 2017-01-29
374 refactor tests for Travis

phnm:
This solves a problem where the here refactored tests would randomly fail only on Travis when implementing feature #271. See #374. Also see comments on PR #440.

### DIFF
--- a/spec/features/participants_spec.rb
+++ b/spec/features/participants_spec.rb
@@ -1,4 +1,5 @@
 require "rails_helper"
+require "object_creation_helper"
 
 RSpec.feature "Event participants overview", :type => :feature do
   before :each do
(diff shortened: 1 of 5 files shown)


spec/features/participants_spec.rb
```diff
@@ -47,48 +47,40 @@
...
     user = FactoryGirl.create(:user)
     profile = FactoryGirl.create(:profile, user: user, last_name: "Otti")
-    application_letter = FactoryGirl.create(:application_letter_accepted, 
+    application_letter = FactoryGirl.create(:application_letter_accepted,
       user: user, event: @event, vegetarian: true, allergic: true)
     user = FactoryGirl.create(:user)
     profile = FactoryGirl.create(:profile, user: user, last_name: "Benno")
-    application_letter = FactoryGirl.create(:application_letter_accepted, 
+    application_letter = FactoryGirl.create(:application_letter_accepted,
       user: user, event: @event)
```
querenker: I know it was already there, but I think this block is very hard to read. If I understand it correctly the purpose is do add some application letters from different users (with different names) to the event. Maybe a factory can be used that generates a user with a profile and a given name? Furthermore the application letters could be generated in a loop which circles through a collection of names (the same collection can then be used below to reduce the magic values)?

querenker (changes requested):

spec/features/participants_spec.rb
phnm: Yep that block creates a bunch of applications with different names (to distinguish them, as this test is about the sorting functionality of the participants table), which requires different users with profiles. Butmost importantly the applications also have different eating habits (vegetarian, allergic etc; as this test is about sorting by those). So you'd need more than a list different names, but also different eating habits. I don't think you can use a factory that, since you need 3 objects of different classes, but I could write a helper method that takes the name, event, and eating habits as parameter and creates the respective user, profile and application letter for it, if you really think that's necessary. I don't think it would be needed anywhere else.

querenker: I think it would still improve readability, but I see your point and it is currently not as bad as I feared initially.

phnm: How about I add a comment to that block, explaining what that block does?
If that's ok and it's the only thing on this PR you'd like changed, I'd appreciate if this PR would be approved, since it's blocking my other PR and that blocks our Kanban board.

querenker (approved):
Otherwise it looks good for me. I would still extract the duplicated code.

Issue #399, opened 2017-01-24
Remove max participants infinity


ekrebs5:
Instead of infinity a large number should be sufficient

bjrne:
So this would consist of removing and adapting the following: (Feel free to edit and adapt)
- [x] database: remove field (event->unlimited)
- [x] remove modification from getter & setter for max_participant
- [x] remove translation for infinity stuff from the locales
- [x] remove checks for infinity in views (especially event show page / applicants overview)
- [x] remove checkbox for infinity from event creation/editing
- [x] remove assignments of max_participants to infinity (e. g. in tests)
- [x] remove it from event params in event controller
- [ ] ?

ekrebs5:
now part of #16

Issue #16, opened 2016-11-10
US_3.1: Accept-All-Button


julisa99:
**As** 
organizer

**I want to**
accept all applicants with an Accept-All-Button if the number of participants is not limited

**in order to**
to be faster

**estimation: 8**

## Tasks

- [x] add accept-all-functionality to button
- [x] only organizers see the button
- [x] the button is only visible if there is enough space
- [ ] remove "infinite numer of participants"-functionality to events
- [ ] remove UI button for infinite participants to event creation
- [ ] remove UI button for infinite participants to event editing

## Acceptance criteria

- The not-limited check box is removed from the application creation page. 

- If there are less or equal applicants than maximum number of participants, than the accept all button should be displayed to organizers visiting the overview page. 

- If the number of applicants is larger than the maximum number of participants, than the button should be hidden. 

- Coaches and applicants never see this button.

- When the button is pressed, all applicants for this workshop should be marked as accepted. 

- The accepts are persistent.

Baschdl:
@tobiduer Only if the number of participants is not limited or also if the number of applications is less than the number of possible participants?

tobiduer:
@Baschdl: Than you for the question. The Accept-All-Button should be displayed in both cases. 

As you might already have recognized, this user story also includes the wish to set the maximum number of participants to "not limited" and the UI should reveal this to the user (e.g. with a button next to the input field)

tobiduer:
**Acceptance criteria**

- [ ] If the maximum number of participants for a workshop is not limited (as defined in the workshop edit view), than the accept all button should be displayed to organizers visiting the overview page. 

- [] If there are less or equal applicants than maximum number of participants, than the accept all button should be displayed to organizers visiting the overview page. 

- [] If the number of applicants is larger than the maximum number of participants, than the button should be hided. 

- [] Coaches and pupils do never see this button.

- [] When the button is pressed, all applicants for this workshop should be marked as accepted.

tobiasduerschmid:
As discussed on Slack, removing the not-limited checkbox is part of this user story. (AC's are updated)

PR #447, opened 2017-02-01
Refs #399: Remove infinity participants (#16)

querenker:


spec/factories/events.rb
```diff
@@ -123,6 +123,12 @@
       end
     end
 
+    trait :with_enough_free_places do
+      after(:build) do |event|
+        event.max_participants = event.application_letters.size + 1
```
Lando-L: Why do you need the addition plus one here? Or is it just for safety reason if there are no application letters or something else?

querenker: Thanks, the plus one is indeed not necessary :) Should now be fixed

bjrne (dismissed):
Migration is missing

spec/factories/events.rb
aloeser: I disagree. If you got 0 application letters yet, you will set max_participants to 0, which is not a valid value

spec/factories/events.rb
```diff
@@ -123,6 +123,12 @@
       end
     end
 
+    trait :with_enough_free_places do
+      after(:build) do |event|
+        event.max_participants = event.application_letters.size
```
aloeser: Besides the missing +1 at the end breaks the tests atm, I'm not really sure if its a good idea to have a trait like this at all. From my understanding, this works only if all applications were added before, e.g. in another trait. So if you put this trait before :with_diverse_open_applications or similar, it will not work as expected. Imo that could bring more problems than benefits for testing

db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170128122851) do
+ActiveRecord::Schema.define(version: 20170131092522) do
```
ekrebs5: maybe I missed it but the migration for this seems to be missing in the pr?

ekrebs5 (changes requested):

spec/factories/events.rb
querenker: Thankfully we have tests to check this :) The last commit is now reverted

spec/factories/events.rb
querenker: Before these changes we could set the max_participants to infinity to ensure that there are enough free spaces. Now we could set a big number and hope that there aren't enough free places, but I think this is even uglier.... Maybe it would be the best if we set this value in the tests by hand?

db/schema.rb
querenker: @bjrne @ekrebs5 Now they should be there

spec/factories/events.rb
ekrebs5: I'm for setting it by hand or making a trait like "adds application and enough free spaces" (because that would imply that it might not be wise to use it with other application traits, like in draft phase and in execution phase would not make sense together)

querenker: Now the value gets set by hand in every test. I think that is currently the least ugly solution.

ekrebs5 (approved):

Baschdl (approved):

Issue #430, opened 2017-01-27
Multiple UI Improvements


hendraet:
**Acceptance Criteria**
- [x] login and register box have the same height 
- [ ] button in "Hochladen-Banner" should be smaller (xs -tag)
- [x] rename label "Über XX Tage" in "Umfang: XX Tage"

tobiasduerschmid:
@hefowe: Could you please check the "Hochladen-Banner" ?

cmfcmf:
I suppose you mean @hendraet?

tobiasduerschmid:
Yes :D

hendraet:
:purple_heart:

PR #460, opened 2017-02-01
Ui improvements

cmfcmf:
#430


config/locales/de.events.yml
```diff
@@ -107,10 +107,10 @@ de:
         other: "Bewerbungsfrist in %{count} Tagen"
       time_span_non_consecutive:
         one: "Über einen Tag mit Unterbrechung"
```
ekrebs5: should there be an Umfang: here?

ekrebs5 (changes requested):

config/locales/de.events.yml
cmfcmf: Thanks! Missed it, fixed.

ekrebs5 (approved):

bjrne:
@cmfcmf It looks like this broke our login page, could you or someone take a look at it?

PR #461, opened 2017-02-01
Randomize first_name of Profiles created by FactoryGirl

Blaidd-Drwg:
This one-line PR reduces false positives when testing sorting methods by randomizing the names created by the Profile factory.

(**Background:** The bug which caused #374 had been in the system for quite some time, but wasn't discovered because the sorting tests passed when they were supposed to fail. The reason for that is the way participant names are created in the Profile factory: in order to make them different from one another, the string 'Karl' is concatenated with the Profile's id. That however means that in most cases sorting by name and sorting by id has the exact same result. Tests checking if participants were correctly sorted by name may pass even if the sorting method didn't do anything at all.)


Fabian-Sommer (approved):

cmfcmf (approved):

Fabian-Sommer (changes requested):
A test is currently failing and needs to be refactored (this change is ok, but has a chance to break that test when multiple persons of the same name are created). See https://travis-ci.org/hpi-swt2/workshop-portal/builds/197377297 )

Blaidd-Drwg:
Oh, it's even a test I wrote myself. How awkward :smile: 
I've changed it to use the full name, which is still unique. I also repeated the tests with a modified factory to double check, no other test relies on the first_name being unique.

Fabian-Sommer (approved):

Issue #427, opened 2017-01-27
Dummy Image For Banner On Event Detail Page


hendraet:
**Acceptance Criteria**
- [ ] banners on event detail page should display an image with colorful shapes

cmfcmf:
Assigned myself now @radscheit.

hendraet:
:neckbeard:

PR #463, opened 2017-02-01
Add colorful shapes instead of placeholder image. (#427)

cmfcmf:
(#427) The shape is randomly generated based on the event name.
image


app/views/events/_show_event_large.html.erb
```diff
@@ -7,7 +7,13 @@
   </div>
   <h1><%= @event.name %></h1>
 </div>
-<%= image_tag "event-detail.png", class: "img-rounded img-responsive center-block m-t-1 m-b-1" %>
+<div id="wsp-event-show-banner"></div>
+<script>
```
nilsthamm, line 11: Ah short comment about the purpose of the script would be good. Something like: "Generate random image, in order to not have the same static image everywhere". So later no one will wonder why this script is necessary.

nilsthamm (changes requested):

nilsthamm (approved):
LGTM, thx

PR #464, opened 2017-02-01
Update production

tobiasduerschmid:
Because of many substantial changes (adding texts and adjusting footer), the production branch should be updated.


bjrne:
@tobiduer Is there something to do here? If not, please use the "request review" feature.

ekrebs5 (approved):

PR #468, opened 2017-02-01
339 3.29 email address profile

Baschdl:
Add the email address to the profile. We don't need extra abilities because only the organizer and admin can view the profiles of other persons.
#339


spec/features/profile_spec.rb
```diff
@@ -137,4 +137,4 @@ def mock_writing_to_filesystem
 
     expect(page).to have_css(".has-error", count: 3)
   end
-end
+end
```
bjrne: Why did you remove the trailing newline here?

bjrne (approved):
Good, but please add trailing newlines in both files 🏒

ekrebs5 (approved):

spec/features/profile_spec.rb
Baschdl: Fixed it

Issue #418, opened 2017-01-25
Event Transition from Selection to Execution


tobiasduerschmid:
**As**
organizer

**I want**
the event to change its [state](233) from selection to transition when the emails for acceptance and rejection are send

**in order to**
be able to send both

# Acceptance criteria

- [x] The organizer is able to send acceptance and rejection mails.

- [x] After sending acceptance or rejection mails, the state is frozen for each application.

- [x] The transition to the execution state takes place after sending both acceptance AND rejection mails 

- [x] The visibility of the buttons is handled according to #244

- [x] Hide the "Zusagen versenden" button, after sending acceptance emails and the "Absagen versenden" button after sending rejection mails

tobiasduerschmid:
Could you please hide the "Zusagen versenden" button, after sending acceptance emails and the "Absagen versenden" button after sending rejection mails?

PR #469, opened 2017-02-01
Refs #418: Event transition from selection to execution

querenker:
- in selection phase, the mail buttons are always visible
- in selection phase, if only kind of mails was sent, the participant selection is locked (aka the radio buttons disappear)
- when both mails have been sent, phase changes from selection to execution
- in execution phase, the mail buttons are never visible 

====================================================
- there are now two boolean flags to indicate whether application rsp. rejection letters have been sent
- lock_application_status was removed as database entry (to get something nearly identical participant_selection_locked can be used)


cmfcmf (approved):
Looks good! Noticed this today while developing the ical attachment feature. 👍

app/models/application_letter.rb
```diff
@@ -46,7 +45,7 @@ def after_deadline?
   # @param none
   # @return [Boolean] true if no status changes are allowed anymore
```
aloeser: Not your code, but a function that checks if a change is allowed returns true when its not?

spec/features/email_spec.rb
```diff
@@ -29,15 +29,16 @@
 
   scenario "logged in as Organizer after sending an Email to the applicants the event application status gets locked" do
     login(:organizer)
-    @event.application_status_locked = false
+    @event.acceptances_have_been_sent = false
```
aloeser: Would it be nicer to have an event in selection phase than setting those values by hand? You might have to check that participant_selection_locked is false at the beginning, tho

spec/features/events_spec.rb
```diff
@@ -152,7 +152,8 @@
 
   scenario "logged in as Organizer I can not change application status with radio buttons if the applications are locked" do
     login(:organizer)
-    @event.lock_application_status
+    @event.acceptances_have_been_sent = true
+    @event.rejections_have_been_sent = true
```
aloeser: same as above, with event in execution phase?
Also, you should propably test that sending either acceptances or rejections alone is enough to lock the participant selection.

Also the test in line 162 (cant mark it) could be updated, since we check for labels now (afaik) when we want to test for radio buttons, not for normal text. It's not that important tho

spec/models/application_letter_spec.rb
```diff
@@ -76,15 +76,17 @@
 
   it "can not be updated if status is changed and application status is locked" do
     application = FactoryGirl.build(:application_letter)
+    application.event.acceptances_have_been_sent = true
+    application.event.rejections_have_been_sent = true
```
aloeser: that seems the same as using an event in execution phase.
Maybe we need another test case, with an event in selection phase & participant selection is locked?
And you should rename the "application status" in the test name

spec/models/application_letter_spec.rb
```diff
@@ -76,15 +76,17 @@
...
     application.status = :rejected
-    application.event.application_status_locked = true
     expect(application).to_not be_valid
   end
 
   it "can be updated if status is changed and application status is not locked" do
     application = FactoryGirl.build(:application_letter_deadline_over)
+    application.event.acceptances_have_been_sent = false
```
aloeser: looks like an event in selection phase

spec/views/events/participants.html.erb_spec.rb
```diff
@@ -52,7 +52,8 @@
   it "displays print badges button (in event execution phase)" do
     @event.published = true
     @event.application_deadline = Date.yesterday
-    @event.application_status_locked = true
+    @event.acceptances_have_been_sent = true
```
aloeser: looks like an event in execution phase

app/models/application_letter.rb
```diff
@@ -46,7 +45,7 @@ def after_deadline?
...
   def status_change_allowed?
-    !event.application_status_locked
+    !event.participant_selection_locked
```
aloeser: I think this function should be changed soon, atm it allows changing the status in both draft and application phase (which is also the reason, why some of our tests are still working that might break otherwise). However, this is not part of this PR, just wanted to point it out

aloeser (changes requested):

spec/features/email_spec.rb
querenker: I think this test should work independent of the current state. After sending an email the participant selection should be locked. You are right, that a emails can only be send in selection phase, but I think that should not be part of the test.

spec/features/events_spec.rb
querenker: Same as above. If the statuses are locked the radio buttons should not be shown (independent of the current phase, but currently the status can only be locked in execution phase). But maybe that will change later?

spec/models/application_letter_spec.rb
querenker: Same as above.

spec/features/events_spec.rb
querenker: But you are right it should be enough if only one flag is logged.

spec/views/events/participants.html.erb_spec.rb
querenker: In this case we should use indeed a factory for an event in execution phase (it is even in the title of the test) :)

app/models/event.rb
```diff
@@ -47,7 +48,9 @@ def participants_by_agreement_letter
     @participants.sort { |x, y| self.compare_participants_by_agreement(x,y) }
   end
 
-
+  def participant_selection_locked
```
aloeser: model test is missing

spec/features/email_spec.rb
```diff
@@ -27,17 +27,30 @@
     expect(page).to have_text(I18n.t('.emails.submit.sending_successful'))
   end
 
-  scenario "logged in as Organizer after sending an Email to the applicants the event application status gets locked" do
+  scenario "logged in as Organizer after sending an acceptance Email to the applicants the event application status for sending acceptances gets locked" do
```
aloeser, line 30: I think this is not what the original test meant. It is still supposed to make sure that you cant change application status after that mail, not that sending an acceptance mail is noticed.

spec/features/email_spec.rb
```diff
@@ -27,17 +27,30 @@
...
+    @event.acceptances_have_been_sent = false
     @event.save
 
     visit event_email_show_path(@event, status: :acceptance)
     fill_in :email_subject, with: "Subject"
     fill_in :email_content, with: "Content"
     click_button I18n.t('.emails.email_form.send')
 
-    expect(Event.find(@event.id).application_status_locked).to be(true)
+    expect(Event.find(@event.id).acceptances_have_been_sent).to be(true)
```
aloeser, line 40: so this should test imo participant_selection_locked, not (only) acceptances_have_been_sent

spec/features/email_spec.rb
```diff
@@ -27,17 +27,30 @@
...
+  end
+
+  scenario "logged in as Organizer after sending an rejection Email to the applicants the event application status for sending rejections gets locked" do
```
aloeser, line 43: same as above

spec/features/events_spec.rb
```diff
@@ -72,7 +72,7 @@
   end
 
   scenario "logged in as Organizer I want to be able to send an email to all accepted applicants" do
-    @event = FactoryGirl.create(:event, :in_selection_phase)
+    @event = FactoryGirl.create(:event, :in_selection_phase, acceptances_have_been_sent: false)
```
aloeser: I think you should check that `acceptances_have_been_sent` is true after sending the email, otherwise we could skip setting it here?

spec/features/events_spec.rb
```diff
@@ -90,7 +90,7 @@
   end
 
   scenario "logged in as Organizer I want to be able to send an email to all rejected applicants" do
-    @event = FactoryGirl.create(:event, :in_selection_phase)
+    @event = FactoryGirl.create(:event, :in_selection_phase, rejections_have_been_sent: false)
```
aloeser: same as above

spec/features/events_spec.rb
```diff
@@ -150,15 +150,18 @@
...
-        expect(page).not_to have_text(I18n.t "application_status.#{new_status}")
+    [[true, true], [true, false], [false, true]].each do |acceptances_have_been_sent, rejections_have_been_sent|
+      @event.acceptances_have_been_sent = acceptances_have_been_sent
+      @event.rejections_have_been_sent = rejections_have_been_sent
+      @pupil = FactoryGirl.create(:profile)
+      @application_letter = FactoryGirl.create(:application_letter, event: @event, user: @pupil.user)
+      visit event_path(@event)
+      ApplicationLetter.statuses.keys.each do |new_status|
+        if new_status != @application_letter.status
+          expect(page).not_to have_text(I18n.t "application_status.#{new_status}")
```
aloeser: I think we are using something like `have_css('label', text: '...')` to check for the radio buttons

spec/models/application_letter_spec.rb
aloeser: Title of the test still needs to be renamed.
Also, shouldnt you somehow make sure that `participant_selection_locked` is actually true? Either by manually checking it, or just by using a factory. Because it works with how `participant_selection_locked` is defined atm, but that might change

spec/models/application_letter_spec.rb
aloeser: Test name should be updated.
Also like above, maybe check that participant selection is actually not locked or use a factory

aloeser (changes requested):

spec/features/email_spec.rb
querenker: Through the new model test you can ensure that if an acceptance mail has been sent the status gets locked. I think a test like this should only test a single functionality.

spec/features/email_spec.rb
querenker: same as above

spec/features/email_spec.rb
querenker: same as above

spec/features/events_spec.rb
querenker: That the value gets changed is already ensured in a different test :) But skipping sounds nice

spec/features/events_spec.rb
querenker: This test was already there only the indentation gets changed. But I can look for it.

spec/models/application_letter_spec.rb
querenker: There is now a model test to check this

spec/features/email_spec.rb
aloeser: I dont think you should rely on a model test for that, how will you notice then you have to change test this too, if the model changes and there is no reference to `participant_selection_locked`?
However, I didnt see that you changed a feature test 2 screens further down, that checks that one of the two is enough to lock it. So its all fine now :)

aloeser (approved):
worked for me locally

Issue #470, opened 2017-02-01
Show apply button for not-logged-in users


tobiasduerschmid:
**As**
user how is not logged in

**I want to**
see the apply button

**in order to**
see, that I can apply directly in the portal

# Tasks

# Acceptance criteria

- All of #286 

- User, that are not logged in, see the button

PR #471, opened 2017-02-01
Fix: Show apply button for guests again (#470).

cmfcmf:
This also renames the ability from "apply" to "view_apply_button".


app/models/ability.rb
```diff
@@ -64,7 +67,7 @@ def initialize(user)
            :view_unpublished, :show_eating_habits, :print_applications_eating_habits, :view_hidden], Event
       can :send_email, Email
       can [:manage, :set_contact_person, :set_notes], Request
-      cannot :apply, Event
+      cannot :view_apply_button, Event
```
bjrne, line 70: Doesn't `organizer` inherit all abilities of `coach`? Or at least shouldn't it?

aloeser: It does, but someone added `:manage` as a permission of Organizers for events. And manage is an alias for all possible actions, so it also allows the organizer `view_apply_button`.
This is a thing that needs refactoring (manage needs to be replaced). I think I already mentioned it in the Deny Apply for Coaches and Organizers PR, but @Lando-L figured out it would be too much things to change, so we decided to make a technical debt issue rather than refactoring it directly in his PR

Lando-L (approved):

PR #472, opened 2017-02-01
Include time period on requests show page

MasterCarl:
I had included the time period in the view spec and developed against that. However, there was no suitable value specified in the factory, resulting in a false positive. I fixed it now @julisa99.


bjrne (approved):
👍 🐨

PR #485, opened 2017-02-02
Use Gem instead of JavaScript implementation of GeoPattern, also use a random HPI color.

cmfcmf:
@hendraet


philipp-bode (approved):
Lookin' fine

PR #486, opened 2017-02-02
389 access rights

AlecSchneider:


cmfcmf:
I don't think this is the best way to do this. I'll suggest some changes once I'm home.

cmfcmf (changes requested):
You should use a devise helper to ensure proper access checks for all methods like done here: https://github.com/hpi-swt2/workshop-portal/blob/dev/app/controllers/application_letters_controller.rb#L2

tobiduer:
Test case for logged-out user is missing

spec/features/event_spec.rb
```diff
@@ -273,20 +273,30 @@
   end
 
   describe "edit page" do
-    it "should preselect the event type" do
+    it "should not be possible to visit as pupil" do
```
tobiasduerschmid, line 289: Test case for logged-out users and coaches are missing. 
it "should not be possible to visit as logged-out user" / it "should not be possible to visit as coach"

tobiasduerschmid (dismissed):
Please add test cases for all requested cases

spec/features/event_spec.rb
JuliusCosmoRomeo: Added the extra test cases, please review again @tobiduer

app/controllers/events_controller.rb
```diff
@@ -6,8 +6,10 @@
 require 'carrierwave'
 
 class EventsController < ApplicationController
+  load_and_authorize_resource
+  skip_authorize_resource :only => [:show,:index,:archive]
```
cmfcmf: Do not skip authorization for these actions. After all, authorization is needed!

app/models/ability.rb
```diff
@@ -48,14 +48,15 @@ def initialize(user)
       can [:create], AgreementLetter
       can [:new, :create], Request
       cannot :view_personal_details, ApplicationLetter, user: { id: !user.id }
+      can [:show, :index, :archive], Event
```
cmfcmf: This must be moved up to line 36 - guests can also do it!

app/models/ability.rb
```diff
@@ -48,14 +48,15 @@ def initialize(user)
...
+      cannot [:edit, :update, :new, :create, :destroy, :accept_all_applicants, :print_applications_eating_habits, :participants_pdf, :participants], Event
```
cmfcmf: This isn't needed - by default you *cannot* do anything.

app/models/ability.rb
```diff
@@ -48,14 +48,15 @@ def initialize(user)
...
     end
     if user.role? :coach
       # Coaches can view Applications and participants for and view, upload and download materials for Event
       can [:view_applicants, :view_participants, :view_material, :upload_material, :print_applications, :download_material], Event
       can [:view_and_add_notes, :show], ApplicationLetter
-      can [:print_applications], Event
       can [:show, :index], Request
-      cannot :view_apply_button, Event
+      cannot [:view_apply_button, :edit, :update, :new, :create, :destroy], Event
```
cmfcmf: Again - only `cannot [:view_apply_button]` is needed here, because it revokes the access given to guests at line 36.

app/models/ability.rb
```diff
@@ -77,6 +80,8 @@ def initialize(user)
     end
     if user.role? :admin
       can :manage, :all
+      can [:edit, :update, :destroy], Event
```
cmfcmf: This is not needed - already granted for organizers in line 73.

app/controllers/events_controller.rb
```diff
@@ -69,13 +72,11 @@ def destroy
 
   # GET /events/1/badges
   def badges
-    authorize! :print_badges, @event
```
cmfcmf: The default authorization check will now check for `badges` instead of `print_badges`. Is this intended?

app/controllers/events_controller.rb
```diff
@@ -143,7 +141,6 @@ def download_agreement_letters
     if not params.has_key?(:selected_participants)
       redirect_to event_participants_url(@event), notice: I18n.t('events.agreement_letters_download.notices.no_participants_selected') and return
     end
-    authorize! :print_agreement_letters, @event
```
cmfcmf: Same here - the check will now be `download_agreement_letters` instead of `print_agreement_letters`.

cmfcmf (changes requested):
I don't think you used the cancancan authorization correctly. Please see my comments and try to correct the cancancan usage.

app/models/ability.rb
JuliusCosmoRomeo: @hendraet told me that it should be possible for guests to see events, so we thought to skip the authorization of these methods.

app/controllers/events_controller.rb
JuliusCosmoRomeo: yes, this is true, forgot to change it.

app/models/ability.rb
cmfcmf: I understand, but I think it's cleaner not to skip it. As a developer, I expect to be able to use the abilities to control the behavior of all of the actions. Skipping authorization just because it's not needed right now feels like introducing technical debt to me.

bjrne:
@cmfcmf Is the feedback addressed? And @AlecSchneider do you still want to finish it or should somebody else take over?

cmfcmf (approved):

cmfcmf:
Looks good now, although I can't test this at the moment.

PR #487, opened 2017-02-02
Make email sending great again!

cmfcmf:
SendGrid should already be configured to send emails with these config changes.

### DIFF
--- a/app/controllers/emails_controller.rb
+++ b/app/controllers/emails_controller.rb
@@ -8,7 +8,7 @@ class EmailsController < ApplicationController
     application_letter_status = get_corresponding_application_letter_status
     @addresses = @event.email_addresses_of_type(application_letter_status)
 
-    @email = Email.new(hide_recipients: true, reply_to: 'workshop.portal@hpi.de', recipients: @addresses.join(','),
+    @email = Email.new(hide_recipients: true, reply_to: Rails.configuration.from_address, recipients: @addresses.join(','),
                        subject: '', content: '')
 
     render :email
--- a/app/controllers/emails_controller.rb
+++ b/app/controllers/emails_controller.rb
@@ -8,7 +8,7 @@ class EmailsController < ApplicationController
     application_letter_status = get_corresponding_application_letter_status
     @addresses = @event.email_addresses_of_type(application_letter_status)
 
-    @email = Email.new(hide_recipients: true, reply_to: Rails.configuration.from_address, recipients: @addresses.join(','),
+    @email = Email.new(hide_recipients: true, reply_to: Rails.configuration.reply_to_address, recipients: @addresses.join(','),
                        subject: '', content: '')
 
     render :email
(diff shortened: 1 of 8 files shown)


aloeser:
Are you sure you dont mix up "from" and "reply to" in some lines? Those can be different (and should be atm)

cmfcmf:
> Are you sure you dont mix up "from" and "reply to" in some lines? Those can be different (and should be atm)

When would these need to be different?

aloeser:
> When would these need to be different?

E.g. when you send all emails with the same mail adress, you still might want different reply-to-adresses, like 1 per workshop/camp, so replies are better organized.
Another reason: you have a strict email provider, that doesnt allow you to send from a @hpi.de adress, when you use a @gmail.com adress.

If you say they are the same all the time, I think we can skip the reply-to completely

cmfcmf:
I don't see how this PR affects changing the reply to address. I only changed hard-coded values which aren't changeable by the end user.

app/controllers/emails_controller.rb
```diff
@@ -8,7 +8,7 @@ def show
     application_letter_status = get_corresponding_application_letter_status
     @addresses = @event.email_addresses_of_type(application_letter_status)
 
-    @email = Email.new(hide_recipients: true, reply_to: 'workshop.portal@hpi.de', recipients: @addresses.join(','),
+    @email = Email.new(hide_recipients: true, reply_to: Rails.configuration.from_address, recipients: @addresses.join(','),
```
aloeser: I think it was already mentioned in another PR(#355) that this value should be in the config, not hardcoded. However we use workshop.portal@gmail.com as address to send mails atm, and the default reply to is workshop.portal@hpi.de.
TLDR: I agree its good to read this value from the config, but this is the wrong config value

cmfcmf:
Why would we ever send emails from @gmail and request to be replied @hpi? I
don't see the use case to change the reply address _for all events at
once_. If a later PR adds support to change the reply address on an event
basis, the code needs to be refactored anyway.

Alexander Löser <notifications@github.com> schrieb am Do., 2. Feb. 2017,
18:32:

Baschdl:
Looks good but I didn't test it local. Did you set the environment variables on heroku?

cmfcmf:
> Looks good but I didn't test it local. Did you set the environment variables on heroku?

Yes.

aloeser (approved):
Worked for me, but it might take a lot of time for the email to be actually delivered

tobiasduerschmid:
@cmfcmf: As you can see on the [email sending page](http://workshop-portal-dev.herokuapp.com/events/7/emails?utf8=%E2%9C%93&status=acceptance&button=)
The reply to address can be specified by the sender in order to enter his / her email address / the email address for managing the event. This is necessary since different events might be managed by different people.

cmfcmf:
> The reply to address can be specified by the sender in order to enter his / her email address / the email address for managing the event. This is necessary since different events might be managed by different people.

Yes, I understand that. The thing is, this PR didn't change anything about that :)

tobiasduerschmid:
Ok. So the email address entered in the "Absender" field will be used as reply-to address? Than I am fine :)

cmfcmf:
> Ok. So the email address entered in the "Absender" field will be used as reply-to address? Than I am fine :)

Yes, of course!

PR #488, opened 2017-02-03
Rename profiles according to their roles and add past-deadline event

tobiasduerschmid:
This PR changes the names of the profiles according to their roles and adds the summer camp as new event used as past-deadline-event. This will improve the experience of the demo.

I could not test if this changes work, because I have no setup of the project, so please check if I did something wrong


tobiasduerschmid:
I am not sure, what I did wrong, but it does not find "event_summer_camp" which I defined in events.rb

https://travis-ci.org/hpi-swt2/workshop-portal/builds/197920436

db/sample_data/events.rb
```diff
@@ -129,3 +129,24 @@ def event_bachlorpodium
       published: true
   )
 end
+
+def event_mintcamp
```
bjrne: Nope, you defined event_mintcamp again.

bjrne:
@tobiduer If you can't test things before opening a PR, please consider doing pair-programming with a developer if you feel like you want to add some code as well. We collectively voted on things that need to be done in order to open a PR in the wiki.

db/sample_data/events.rb
tobiasduerschmid: Oh. I forgot to save the file in sublime -.-

tobiasduerschmid:
@bjrne wishes to create a new US that developers should implement instead of POs.

PR #504, opened 2017-02-03
418 3 27 event transistion selection execution small addition

ekrebs5:
sending mail buttons disappear after use


spec/views/events/show.html.erb_spec.rb
```diff
@@ -211,6 +211,24 @@
     expect(rendered).to_not have_link(t(:show_participants, scope: 'events.participants'))
   end
 
+  it "does not display send acceptances button after acceptances have been sent in selection phase" do
+    @event = assign(:event, FactoryGirl.create(:event, :in_selection_phase))
```
querenker: I think it is not important if the event is in selection phase. If the acceptances have been sent then the button should not be displayed independent of the current phase. On the other hand I think it is important that there are some accepted applications for the event. Otherwise the button could be hidden just because there are no (accepted) applications.

spec/views/events/show.html.erb_spec.rb
```diff
@@ -211,6 +211,24 @@
...
+    @event.acceptances_have_been_sent = true
+    @event.save
+    sign_in(FactoryGirl.create(:user, role: :organizer))
+    render
+    expect(rendered).to_not have_button(t(:sending_acceptances, scope: 'events.applicants_overview'))
+  end
+
+  it "does not display send acceptances button after acceptances have been sent in selection phase" do
+    @event = assign(:event, FactoryGirl.create(:event, :in_selection_phase))
```
querenker: See above

spec/views/events/show.html.erb_spec.rb
```diff
@@ -211,6 +211,24 @@
     expect(rendered).to_not have_link(t(:show_participants, scope: 'events.participants'))
   end
 
+  it "does not display send acceptances button after acceptances have been sent in selection phase" do
+    @event = assign(:event, FactoryGirl.create(:event, :in_selection_phase))
+    @event.acceptances_have_been_sent = true
```
querenker: I think this flag could be set by factory creation, but I think it doesn't matter.

querenker (changes requested):
Otherwise it looks good for me

spec/views/events/show.html.erb_spec.rb
aloeser: > I think it is not important if the event is in selection phase.

I really disagree with that. The buttons to send acceptance and rejection emails are (atm) only displayed in selection phase, so using an event that is not in selection phase does not make sense at all, because the buttons shouldnt be displayed anyway.

>  On the other hand I think it is important that there are some accepted applications for the event. Otherwise the button could be hidden just because there are no (accepted) applications.

that could happen at some point in the future, but imo its not relevant yet, because the current implementation checks whether a mail was sent, not if there are receivers waiting for one.

In general, I dont see the point in not using the phases in tests. Not using them to specify additional values makes it only easier to get such false positives like this one would be, when the test isnt green because the code is working, but because the code to test isnt even executed

querenker: > I really disagree with that. The buttons to send acceptance and rejection emails are (atm) only displayed in selection phase, so using an event that is not in selection phase does not make sense at all, because the buttons shouldnt be displayed anyway.

> In general, I dont see the point in not using the phases in tests. Not using them to specify additional values makes it only easier to get such false positives like this one would be, when the test isnt green because the code is working, but because the code to test isnt even executed

Suppose you add some new states between selection and execution state, you have to copy every test like that to still ensure that the emails can only be sent once. I think its not so much fun and likely that you forget some test cases. I agree with you that currently the flag is only relevant in selection phase, because before there shouldn't be any option to set this flag and in execution phase the button will never be displayed, but in the future? I don't know..... You are right that leaving out the phase could produce some trouble, because in that case the state depends on the default implementation in the factory. Instead of testing any phase there should be tested every phase, but I think currently it is not so easy to access every possible event phase and it isn't done everywhere else, so I think it does not have to part of this PR.

querenker: > that could happen at some point in the future, but imo its not relevant yet, because the current implementation checks whether a mail was sent, not if there are receivers waiting for one.

You are right, for any reason I thought that it is not possible to send acceptance mails if there are no accpeted applicants. But currently there isn't a check like that so it is fine

querenker (approved):

timbasel (approved):
Looks good

PR #455, opened 2017-02-01
431 remove button chaos on event creation page

nstrelow:
refs: #431 

remove kind
hidden now is displayed as choice between "public" and "private" for the user

PR #508, opened 2017-02-03
242 3.25 application canceled state refactoring

aloeser:
This PR contains some refactorings/preparations for actual userstory, in order to keep the PR at a (more) reasonable size
- add status_notification_sent-flag to application letters, which keeps track whether an email about the latest status change of their application has been sent to the applicants
- introduce an additional application letter state: canceled, but dont show it as a radio button etc
- some smaller refactorings where needed


spec/features/events_spec.rb
```diff
@@ -166,6 +167,42 @@
     end
   end
 
+  scenario "logged in as Organizer I can send acceptance and then rejection emails and by that change the status notification flag" do
+    login(:organizer)
+    event = FactoryGirl.create(:event_with_accepted_applications, :in_selection_phase)
+
+    event.application_letters.each do |application|
```
timbasel: Can this get moved into the Factory, it shouldn't really be part of the test

spec/features/events_spec.rb
```diff
@@ -166,6 +167,42 @@
...
+    end
+  
+    applications = event.application_letters.select { | application_letter | application_letter.status == 'accepted' }
+    visit event_path(event)
+    click_button(I18n.t('events.applicants_overview.sending_acceptances'))
+    expect(page).to have_current_path(event_email_show_path(event_id: event.id), only_path: true)
+    fill_in :email_subject, with: "Subject Accepted"
+    fill_in :email_content, with: "Content Accepted"
+    click_button I18n.t('.emails.email_form.send')
+    expect(applications.size).to be > 0
```
timbasel: Shouldn't this be checked before emails are send?

spec/features/events_spec.rb
```diff
@@ -166,6 +167,42 @@
...
+    end
+
+    applications = event.application_letters.select { | application_letter | application_letter.status == 'rejected' }
+    visit event_path(event)
+    click_button(I18n.t('events.applicants_overview.sending_rejections'))
+    expect(page).to have_current_path(event_email_show_path(event_id: event.id), only_path: true)
+    fill_in :email_subject, with: "Subject Rejected"
+    fill_in :email_content, with: "Content Rejected"
+    click_button I18n.t('.emails.email_form.send')
+    expect(applications.size).to be > 0
```
timbasel: see above

timbasel (changes requested):

spec/models/event_spec.rb
```diff
@@ -279,4 +286,36 @@
       expect(event.participant_selection_locked).to eq(acceptances_sent || rejections_sent)
     end
   end
+
+  it "is in draft phase" do
+    event = FactoryGirl.build(:event, :in_draft_phase)
```
ekrebs5: these are the same test as 249-267, right?

ekrebs5 (changes requested):

spec/features/events_spec.rb
aloeser: The default value is already `false`, this block is only here to make sure this test has still a meaning in case the factory's default value switches to true

spec/features/events_spec.rb
aloeser: Maybe, why?
The reason why I wrote this line is the loop below.
```
some_array.each do
  expect(something)
end
```
Such a block is pretty pointless, if you are not sure that the loop is actually executed

timbasel: yes, but the problem is if application is empty, than no email will be send and the status will not be updated

aloeser: This line was mainly introduced because I used where instead of select before, which sometimes returned in 0 instead of the 5 expected applications, allthough they do exist. So I dont think its really needed, but I can move it up 1 line?

spec/features/events_spec.rb
timbasel: ok, but I think than you should a trait or another option that does not change.

ekrebs5 (approved):

spec/features/events_spec.rb
timbasel: yeah, that was what I would suggest, to move it directly below the select statement

app/controllers/emails_controller.rb
```diff
@@ -29,24 +29,27 @@ def submit
...
         @email.send_email_with_ical @event
       else
         @email.send_email
       end
 
-      if get_email_template_status == :acceptance
+      if status == :acceptance
         @event.acceptances_have_been_sent = true
-      elsif get_email_template_status == :rejection
+        @event.set_status_notification_flag_for_applications_with_status(application_letter_status)
```
querenker: I think this line could be extracted from the if-else blocks?

app/models/application_letter.rb
```diff
@@ -25,13 +25,22 @@ class ApplicationLetter < ActiveRecord::Base
...
+  enum status: {accepted: 1, rejected: 0, pending: 2, alternative: 3, canceled: 4}
   validates :status, inclusion: { in: statuses.keys }
 
 
+  # Returns an array of selectable statuses
+  #
+  # @param none
+  # @return [Array <String>] array of selectable statuses
+  def self.selectable_statuses
+    return ["accepted","rejected","pending","alternative"]
```
querenker: I know that we have no real coding standards, but I think in most places the return statement is not used if it is not neccessary?

app/models/application_letter.rb
```diff
@@ -98,6 +123,14 @@ def status_cannot_be_changed
     end
   end
 
+  # Validator for status_change_allowed?
+  # Adds error
+  def status_notification_sent_cannot_be_changed
+    unless status_notification_sent_change_allowed?
+      errors.add(:event, "Das Status-Benachrichtungsflag kann noch nicht gesetzt werden") #TODO
```
querenker: Because of the #TODO if think you already know, that the string have to be localised.

Baschdl:
Travis outputs a DEPRECATION WARNING. Maybe this is the cause for the strange message from coveralls. Did you add this assertion?
`DEPRECATION WARNING: The assertion was not run because of an invalid css selector.
unexpected '"event[hidden]"' after '[' (called from block (3 levels) in <top (required)> at /home/travis/build/hpi-swt2/workshop-portal/spec/views/events/new.html.erb_spec.rb:17)
DEPRECATION WARNING: The assertion was not run because of an invalid css selector.
unexpected '"event[published]"' after '[' (called from block (3 levels) in <top (required)> at /home/travis/build/hpi-swt2/workshop-portal/spec/views/events/new.html.erb_spec.rb:18)`

timbasel (approved):

app/models/application_letter.rb
```diff
@@ -46,7 +55,21 @@ def after_deadline?
   # @param none
   # @return [Boolean] true if status changes are allowed
   def status_change_allowed?
```
querenker, line 57: Currently the purpose of this method isn't that clear for me. I think a test could help here ;)

Baschdl:
Mhh, you didn't change this file, strange.

timbasel:
@Baschdl This warning is older, I got it on my PR too and there @coveralls worked

app/models/application_letter.rb
```diff
@@ -46,7 +55,21 @@ def after_deadline?
...
+    else
+      true
+    end
+  end
+
+  # Checks if it is allowed to set the status_notification_sent flag
+  #
+  # @param none
+  # @return [Boolean] true if it can be changed
+  def status_notification_sent_change_allowed?
```
querenker, line 71: I know currently this method is very simple and a test would seeming a litttle bit overkill, but I think there should be one nevertheless?

Baschdl:
Yes, PR #455 introduced it

app/models/event.rb
```diff
@@ -150,12 +150,23 @@ def accept_all_application_letters
     end
   end
 
+  # Sets the status_notification_sent flag for all application letters of the given type
+  #
+  # @param status [Type] the desired application status the flag should be set for
+  # @return none
+  def set_status_notification_flag_for_applications_with_status(status)
+    applications = application_letters.select {|application| application.status == status.to_s}
+    applications.each do |application_letter|
```
querenker, line 166: Instead of iterating over the collection I think you could use map! to call a method on every object in a collection. Just a little note. This way it also works and it isn't that bad.

spec/features/events_spec.rb
```diff
@@ -166,6 +167,42 @@
...
+    expect(page).to have_current_path(event_email_show_path(event_id: event.id), only_path: true)
+    fill_in :email_subject, with: "Subject Accepted"
+    fill_in :email_content, with: "Content Accepted"
+    click_button I18n.t('.emails.email_form.send')
+    applications.each do |letter|
+      letter.reload
+      expect(letter.status_notification_sent).to be true
+    end
+
+    applications = event.application_letters.select { | application_letter | application_letter.status == 'rejected' }
```
querenker: I thin this block is very similar to the block above. Maybe the code duplication can be reduced?

spec/views/events/show.html.erb_spec.rb
```diff
@@ -76,6 +76,18 @@
     expect(rendered).to have_css("td", :text => @application_letter.user.profile.age_at_time(@event.start_date))
   end
 
+  it "logged in as organizer it renders radio buttons for accept reject pending and alternative, but not canceled in selection phase" do
+    sign_in(FactoryGirl.create(:user, role: :organizer))
+    @event = assign(:event, FactoryGirl.create(:event, :with_diverse_open_applications, :in_selection_phase))
+    @application_letters = @event.application_letters #TODO I couldnt find a assign(:application_letters), still this gives the view access to it.
```
querenker: TODO

querenker (changes requested):

app/models/application_letter.rb
aloeser: This functions is used for validations, and we never tested our validation functions afaik.
Also, there are model tests in this PR which cover that (via validations) :)

app/models/application_letter.rb
aloeser: Is that really needed for this PR? There is another non-localized string as error message in this file, so it needs to be fixed (later) anyway, so I found no example and the things i tried didnt work.

app/models/application_letter.rb
aloeser: same as `status_change_allowed?`, I dont think so

app/models/application_letter.rb
querenker: I can understand that methods that for example simply negate a value (like this method before) doesn't have a test (even if in my opinion they should have), but this method isn't that easy to understand for me and I think alone for this reason a test would be good?

app/models/application_letter.rb
querenker: Unfortunately there are still non localized strings in the code, but I think new code should be localized anyway. I think it shouldn't be that complicated, but I will check this.

spec/features/events_spec.rb
querenker: I created a PR on your PR for this: https://github.com/hpi-swt2/workshop-portal/pull/518

app/models/application_letter.rb
querenker: There is now a PR on your PR for this: https://github.com/hpi-swt2/workshop-portal/pull/517

app/models/application_letter.rb
aloeser: Its a helper for validation functions, and we have (and introduce) a lot of validation tests already. I dont think we should test helpers separatly

querenker (approved):
I still think that helper functions should be tested too. At least for me (maybe I am simply bad at debugging) it is very beneficial if a test can point me to the broken code and this encloses helper functions especially if they are kind of complex (the test could even help to document this code). But you are right that in most cases there aren't tests for helper functions. I think that is a very bad argument, but you get my approve nonetheless.

aloeser:
We do have tests especially for the function you talk about, and introduced some more. They will directly point you to the wrong code, the only difference is that they test validations for the status, and not directly call the helper :)

Issue #500, opened 2017-02-03
Adjust Sample Data


tobiasduerschmid:
**As**
PO 

**I want to**
have suitable sample data

**in order to**
support the presentation

# Acceptance criteria

- The @hpi.de dummy accounts have the german translation of the role as last name

- Instead of using the Gongakrobatik event twice, a new event (HPI Sommercamp) should be created, whose deadline is over.

bjrne:
If it helps, have a look at the branch `adjust_sample_data`.

tobiasduerschmid:
@nilsthamm: The build of your PR is still broken. #519

tobiasduerschmid:
Can you add line breaks for the event description?

http://workshop-portal-dev.herokuapp.com/events/7

nilsthamm:
yes, i guess it would also be useful for the other events?!

hendraet:
:metal:

PR #519, opened 2017-02-04
Adjust sample data

nilsthamm:
refs #500


db/sample_data/events.rb
```diff
@@ -124,3 +124,24 @@ def event_bachlorpodium
...
+      max_participants: 25,
+      organizer: 'HPI Schülerklub',
+      knowledge_level: 'Fortgeschrittene',
+      date_ranges: [date_range_summer_camp],
+      application_deadline: Date.tomorrow,
+      application_status_locked: false,
```
cmfcmf: This field has recently been split into two and is the reason why the tests fail.

nilsthamm: thank you, i will fix it

tobiasduerschmid:
@nilsthamm: The build on this branch is still broken.

db/sample_data.rb
```diff
@@ -67,8 +68,7 @@ def add_sample_data
   [events, users, profiles, application_letters, requests, agreement_letters].each do |models|
     save_models(models)
   end
-
-  # set deadline to past to work around validation of application letters
```
bjrne: Why did you remove this comment? I think it's important to understand why we already save things here.

bjrne (approved):
Looks good now. I added two very small commits, please have a look at them before merging. Otherwise this looks good 🔪

nilsthamm:
@bjrne thanks for the review, i think i deleted the comment by accident...

PR #523, opened 2017-02-04
242 3.25 use application canceled state

aloeser:
Display buttons in execution phase on events#show and application_letters#show to

- accept alternative applications, if enough free places
- cancel accepted applications if they were informed about their acceptance
- display some fancy envelope glyphicon if applicants were not informed about their acceptance


ekrebs5 (approved):

spec/factories/events.rb
```diff
@@ -217,6 +217,46 @@
...
+        accepted_application_letters_count 1
+        rejected_application_letters_count 1
+        alternative_application_letters_count 1
+        canceled_application_letters_count 1
+      end
+      organizer "Workshop-Organizer"
+      knowledge_level "Workshop-Knowledge Level"
+      application_deadline Date.current
+
+      after(:create) do |event, evaluator|
```
querenker: Is there a reason to use `create` instead of `build`? With `create` I think the line 252 with the `event.save!` could be even removed.

querenker (approved):
Looks good for me. Only one little question, but it is OK nevertheless.

querenker:
The new change looks good to me as well.

spec/factories/events.rb
```diff
@@ -217,6 +217,63 @@
...
+        create_list(:application_letter_rejected, evaluator.rejected_application_letters_count, event: event)
+        create_list(:application_letter_alternative, evaluator.alternative_application_letters_count, event: event)
+        create_list(:application_letter_canceled, evaluator.canceled_application_letters_count, event: event)
+        event.published = true
+        event.application_deadline = Date.yesterday
+        event.acceptances_have_been_sent = true
+        event.rejections_have_been_sent = true
+      end
+
+      trait :with_no_status_notification_sent_yet do
```
timbasel: remove _yet to make it consistent with other test

timbasel (approved):
A small little consistency change, but the rest looks good, so I approve even now

Issue #495, opened 2017-02-03
Test case for main workflows


tobiasduerschmid:
**As**
PO

**I want to**
ensure, that the workflows that will be shown during the demo works correctly all the time

**in order to**
carefree focus on reviewing new features and coordinating user stories. 

# Acceptance criteria 

- The [main workflows](https://github.com/hpi-swt2/workshop-portal/wiki/Klickpfad) are ensured in an automated acceptance test for this concrete control flow.

querenker:
I don't know how good it is, but for me it looks promising?
https://github.com/LRDesign/rspec-steps

PR #524, opened 2017-02-04
#495 [1] test main workflows

bjrne:
The first three steps of the click path. We decided to use (and add) id's where possible in order to test the flow and not the existence of certain buttons or links.


config/locales/de.yml
```diff
@@ -47,10 +47,9 @@ de:
         remember_me: "Eingeloggt bleiben"
   midnight: "Mitternacht"
   start_page:
-    welcome_to: "Herzlich willkommen im"
+    welcome_to: "Herzlich Willkommen im"
     workshop_portal_hpi_html: "Workshop<div class='row spacer-15'></div>Portal des HPI."
     register_now: "Registrieren"
-    welcome: "Herzlich Willkommen"
```
MasterCarl, line 53: Good change, though not necessarily in scope of this story, I think

spec/features/interactions/standard_interaction_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  it 'should let an organizer log in' do
+    #TODO: Use id instead of visible string, currently id is missing for login button
+    click_link I18n.t('navbar.login')
+    login_with_credentials(user_organizer.email, user_password)
+    expect(page).to_not have_text I18n.t('devise.failure.inactive')
+    expect(page).to_not have_text I18n.t('devise.failure.invalid')
+    expect(page).to_not have_text I18n.t('devise.failure.not_found_in_database')
+    expect(page).to_not have_text I18n.t('devise.failure.locked')
+    expect(page).to have_text I18n.t('start_page.welcome_to')
```
MasterCarl: Wouldn't it be better to check for the `sessions.signed_in` message instead?
If you do want to leave it that way, I would include `shows the start page` in the test case description.

MasterCarl (changes requested):
Looks good in general

config/locales/de.yml
bjrne: This isn't used anywhere, we now use the one 4 lines above that. I thought it'd get lost if I didn't remove it now

spec/features/interactions/standard_interaction_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  it 'should let organizer click the new event button' do
+    click_link I18n.t('navbar.events')
+    click_link 'new_event'
+  end
+
+  it 'should let organizer fill out event creation page' do
+    #TODO: Add radio button choice
```
ekrebs5: Are you sure you want these todos here? Isn't this part of the "first three steps"?

bjrne: That's right, I forgot to mention it. I think it is impossible to select them with capybara in their current implementation. Since the default values correlate with the stories needs I left it out first while seeking for help now.

@nstrelow From what I saw, these are two radio buttons that don't get clicked, instead their class changes from `btn btn-default` to `btn btn-default active`, right? Do you have any idea on how to get this accomplished in capybara in a beautiful way?

ekrebs5 (commented):
Looks good in general.
But you still have 3 TODOs in the tests.
Are those on purpose? (as in you want to add them in a later pr for this issue)

bjrne:
@ekrebs Yes, that's the case. We couldn't figure out how to do this, but the general test of creating an event shouldn't be hindered too much. In order to proceed and test the more important stuff I decided to leave it out for now while noting a missing check there.

ekrebs5 (approved):

MasterCarl (approved):

Issue #525, opened 2017-02-05
Forced to send rejections emails even when no applicants were rejected


Blaidd-Drwg:
**As**
organizer 

**I want**
the event to switch to execution, event if no applicant was rejected

**in order to**
ensure correctness of the system.

# Dependencies

- Can interfere with #526 (Rejection Mail for alternatives)

# Acceptance criteria

- All of #418 

- When the organizer accepts all applicants and send the acceptance mail, then the event should switch to [execution state](https://github.com/hpi-swt2/workshop-portal/issues/233)

# Notes

The phase of the event doesn't change to `execution` if no rejection emails have been sent. As a consequence, the participant list button is not displayed.

![screenshot from 2017-02-05 01-25-51](https://cloud.githubusercontent.com/assets/6051384/22622729/2ce6abac-eb43-11e6-96f8-9c04c876ae7d.png)

After sending a "rejection email" to a bogus address, the event switches to `execution` and I can go back to organizing stuff.

![screenshot from 2017-02-05 01-26-37](https://cloud.githubusercontent.com/assets/6051384/22622758/4a2386ee-eb44-11e6-997e-96fc74fa402b.png)

@hpi-swt2/product-owner-2016

tobiasduerschmid:
@ekrebs5: It works well. Currently, Nachrücker are not handled as rejected applicants. This means, that when I do not have rejected applicant, but Nachrücker, the "Absagen verschicken" Button can not be pushed. I guess, you will implement this in #526?

ekrebs5:
yes, I think that would be part of #526

PR #527, opened 2017-02-05
#525 no longer send rejection mails to non-existing people

ekrebs5:
#525

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -193,6 +193,14 @@ class Event < ActiveRecord::Base
     application_letters.where(status: ApplicationLetter.statuses[:accepted]).count
   end
 
+  # Returns the number of currently rejected applications of the event
+  #
+  # @param none
+  # @return [Int] for number of rejected applications
+  def compute_number_of_rejections_without_mail
+    application_letters.where(status: ApplicationLetter.statuses[:rejected], status_notification_sent: false).count
+  end
+
   # Returns the current state of the event (draft-, application-, selection- and execution-phase)
   #
   # @param none
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -198,7 +198,7 @@ class Event < ActiveRecord::Base
   # @param none
   # @return [Boolean] true if there are rejected participants without status notification sent
   def has_rejected_participants_without_status_notification?
-    application_letters.where(status: ApplicationLetter.statuses[:rejected], status_notification_sent: false).count > 0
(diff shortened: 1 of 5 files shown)


spec/features/email_spec.rb
```diff
@@ -40,6 +40,20 @@
     expect(Event.find(@event.id).acceptances_have_been_sent).to be(true)
   end
 
+  scenario "logged in as Organizer after sending an acceptance Email to the applicants of an event which has no rejected applications the event application status for sending rejections gets locked" do
+    @event = FactoryGirl.create(:event_with_accepted_applications,
+                                rejected_application_letters_count: 0,
```
aloeser, line 45: you could use :in_selection_phase here, not sure if we really need it tho

app/models/event.rb
```diff
@@ -193,6 +193,14 @@ def compute_occupied_places
     application_letters.where(status: ApplicationLetter.statuses[:accepted]).count
   end
 
+  # Returns the number of currently rejected applications with no mail sent of the event
+  #
+  # @param none
+  # @return [Int] for number of rejected applications
+  def compute_number_of_rejections_without_mail
```
aloeser: not wrong, but I think what we want to check is if there are rejected participants at all.
maybe we should rename the function to has_rejected_participants and perform the > 0 check in here?

spec/features/email_spec.rb
ekrebs5: the other test don't have it
(same reasoning, phase those not effect email sending once you are on the email site)

unless we find a convention for this (see the discussion alex started on our slack), I would leave it like those

app/models/event.rb
```diff
@@ -193,6 +193,14 @@ def compute_occupied_places
     application_letters.where(status: ApplicationLetter.statuses[:accepted]).count
   end
 
+  # Returns whether there are rejected applications with no status notification sent of the event
+  #
+  # @param none
+  # @return [Boolean] true if there are rejected participants without status notification sent
+  def has_rejected_participants_without_status_notification?
+    application_letters.where(status: ApplicationLetter.statuses[:rejected], status_notification_sent: false).count > 0
```
querenker: I think you can use [exists?](http://apidock.com/rails/ActiveRecord/Base/exists%3F/class). I think it is a little bit better to read and with many applications_letters it is lots faster.

querenker (changes requested):
Otherwise it looks good to me

querenker (approved):

aloeser (approved):
worked for me

PR #529, opened 2017-02-05
[PR on feature branch] 242_3.25 fix failing test

querenker:
It behaves like expected. If you accept an alternative and there are no alternatives left but free places available you can accept rejected applicants. Therefore a button with this label is still available.

### DIFF
--- a/spec/features/events_spec.rb
+++ b/spec/features/events_spec.rb
@@ -184,7 +184,7 @@ RSpec.feature "Event application letters overview on event page", :type => :feat
 
   scenario "logged in as Organizer I can accept alternative applications (execution phase)" do
     login(:organizer)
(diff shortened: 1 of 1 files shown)


aloeser (commented):
I dont think this is what is wanted, since it only fixes the symptom, not the problem. Even if I have an rejected application (thats why I put the 1 there explicitly), Line 193 shouldnt fail. This is because in the expect also the href is given, and the rejected application should have a different URL, so it shouldnt be relevant for that expect.

aloeser:
I found the problem: In line 193 I didnt put the () around the string for I18n.t, which caused the href attribute to be counted towards I18n.t . So it was only checked for the name of the link, and as OP described there is another link with the same text, but different href...
Its already fixed on the branch, so I close this PR

PR #531, opened 2017-02-05
242 3.25 accept rejected without alternatives

aloeser:
This PR contains

- the last requirement for #242  - the possibility to accept rejected applications, when there are enough free places and no alternative applications
- some small refactorings/tests that were made more precise because of the introduced changes


ekrebs5 (approved):

querenker (approved):
Looks good to me 👍

PR #532, opened 2017-02-06
fix event labels

### LINKED ISSUE #499: Fix Labels For Events
**Acceptance Criteria**
- [x] when application deadline is today there should be a label: "Bewerbungsfrist heute"
- [x] when event is only one day long the label should say: "Umfang 1 Tag"

phnm:
closes #499

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -261,8 +261,7 @@ class Event < ActiveRecord::Base
   # @return string containing the label or nil
   def application_deadline_label
     days = (application_deadline - Date.current).to_i
-    return I18n.t('events.notices.deadline_approaching', count: days) if days <= 7 and days > 0
-    return I18n.t('events.notices.deadline_approaching.today') if days == 0
+    return I18n.t('events.notices.deadline_approaching', count: days) if days <= 7 and days >= 0
   end
 
   # Uses the start date to determine whether or not this event is in the past (or more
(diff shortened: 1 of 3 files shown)


config/locales/de.events.yml
```diff
@@ -117,11 +117,12 @@ de:
       deadline_approaching:
         one: "Bewerbungsfrist in einem Tag"
         other: "Bewerbungsfrist in %{count} Tagen"
+        today: "Bewerbungsfrist heute"
```
Blaidd-Drwg: Can't I18n also do `zero`? That would allow you to simplify `Event#application_deadline_label`

app/models/event.rb
```diff
@@ -284,7 +285,7 @@ def duration_label
 
     if date_ranges.size > 1
       I18n.t('events.notices.time_span_non_consecutive', count: days)
-    elsif days > 1
+    elsif days >= 1
```
Blaidd-Drwg: The `end_date` can't be before the `start_date`, so this will always be true.

config/locales/de.events.yml
phnm: It seems so. Didn't know that, changed it so.

Blaidd-Drwg (approved):
Looks absolutely amazing to me. 10/10 PR

Issue #526, opened 2017-02-05
Rejection mail for alternatives


tobiasduerschmid:
**As**
organizer 

**I want to**
send rejection mails to alternatives (Nachrücker)

**in order to**
handle them like rejected applicants.

# Dependencies

- Can interfere with #525 (Transition from Selection to Execution when no rejection mails need to be send)

# Acceptance criteria

- When the organizer pushes "Absagen verschicken" button, the list of rejected applicants and alternatives are used as receivers of the email.

- When there are no rejected applicants, but Nachrücker, it should be possible to send letters of rejection (See #525 and #418)

ekrebs5:
For developers:
- update the has_rejected_participants_without_status_notification? function in event to also include alternatives

PR #536, opened 2017-02-06
send rejection mail also to alternatives #526

ekrebs5:
Rejection mails are also send to alternatives

old functions that were adjusted:
- if you send acceptions, it switches directly to execution if there are no rejection or alternatives
- the alternatives also get the notifiction send flag

### DIFF
--- a/app/controllers/emails_controller.rb
+++ b/app/controllers/emails_controller.rb
@@ -7,7 +7,7 @@ class EmailsController < ApplicationController
     @templates = EmailTemplate.with_status(get_email_template_status)
     application_letter_status = get_corresponding_application_letter_status
     @addresses = @event.email_addresses_of_type_without_notification_sent(application_letter_status)
-    if (get_email_template_status == :rejection) && (@event.has_alternative_participants_without_status_notification?)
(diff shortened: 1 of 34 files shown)


app/controllers/emails_controller.rb
```diff
@@ -49,11 +52,12 @@ def send_application_result_email
       @event.set_status_notification_flag_for_applications_with_status(application_letter_status)
```
frederike-ramin: Maybe this should be within the if clause. If you send emails with a status other than acceptance/rejection, you don't want the flag to be updated.

app/controllers/emails_controller.rb
```diff
@@ -49,11 +52,12 @@ def send_application_result_email
...
       if status == :acceptance
         @event.acceptances_have_been_sent = true
-        if not @event.has_rejected_participants_without_status_notification?
+        if not @event.has_rejected_or_alternative_participants_without_status_notification?
```
frederike-ramin: If you just replace that by a simple or, you can save up the whole function.

frederike-ramin (changes requested):

timbasel (approved):
Looks good to me

frederike-ramin (approved):

Issue #511, opened 2017-02-04
XSS attack via custom MarkdownRenderTruncate


cmfcmf:
# Steps to reproduce
- login as organzier or admin
- edit the event description and include malicious content, i.e. 
```
# Hallo
<script>alert('Get cheap drugs from http://hpi.de.')</script>
```
- go to http://workshop-portal-dev.herokuapp.com/events
- enjoy :) 

# Further thoughts
While this currently only affects organizers and admins, it could also affect other people if the [`MarkdownRenderTruncate`](https://github.com/hpi-swt2/workshop-portal/blob/d7e3194b7788c1c377a2ca7a8e81f527efeb5ec6/app/helpers/application_helper.rb#L6) is ever used for user / guest input.
Also, I was really disappointed when I found out that the session cookie is `httponly`, so I couldn't steal it :(
Of course, you can still use the JavaScript injection to do funky stuff.

@hpi-swt2/product-owner-2016

LeaGe:
i would like to try fixing this. But I'll probably need help. @cmfcmf do you think, we could do pair programming on this?

cmfcmf:
> i would like to try fixing this. But I'll probably need help. @cmfcmf do you think, we could do pair programming on this?

Glad to hear you want to try, looks like I won't be in Potsdam that much to do pair programming next week though 😕

PR #539, opened 2017-02-07
511 xss attack via custom markdown render truncate

hefowe:
Refs #511.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -5,11 +5,11 @@ require 'redcarpet/render_strip'
 # except paragraphs
 class MarkdownRenderTruncate < Redcarpet::Render::Base
   def paragraph(text)
-    text + ' '
+    CGI::escapeHTML(text) + ' '
   end
 
   def link(link, title, content)
-    content
+    CGI::escapeHTML(content)
   end
 end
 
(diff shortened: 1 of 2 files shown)


hefowe:
@cmfcmf is this fix good enough? Is there still a way to bypass it?

cmfcmf:
> @cmfcmf is this fix good enough? Is there still a way to bypass it?

I'm afraid this isn't the correct fix. You now strip all the HTML even if the `Redcarpet::Render::HTML` renderer is used. But we want that HTML.

hefowe:
@cmfcmf you are right.  Which tags should be allowed? Or can we just strip the html from the index event page? I think we do not need the html on this page?

cmfcmf:
> Or can we just strip the html from the index event page?

Well, this is what the `MarkdownRenderTruncate` was initially made for. I think it would be sufficient to call `CGI::escapeHTML(...)` on the `text` and `content` variables in here: https://github.com/hpi-swt2/workshop-portal/blob/d7e3194b7788c1c377a2ca7a8e81f527efeb5ec6/app/helpers/application_helper.rb#L6

hefowe:
@cmfcmf Thanks for your help!

nilsthamm (approved):

Issue #513, opened 2017-02-04
Remote code execution vulnerability


cmfcmf:
# Steps to reproduce
- **Don't do this at Heroku!**
- login as organizer, admin, coach
- got to http://localhost:3000/events/2?sort=destroy!
- You just deleted the profiles of all the applicants. Nicely done :)

# Further thoughts
The current implementation in the [application overview helper](https://github.com/hpi-swt2/workshop-portal/blob/dev/app/helpers/applicants_overview_helper.rb#L16-L20) allows the user to call **any(!)** methods on the user profiles of all users who applied for the event.

@hpi-swt2/product-owner-2016

PR #549, opened 2017-02-07
Removed vulnerability by checking for valid methods

philipp-bode:
I removed the edge case of sorting by applicant_age_when_event_starts since the order should remain the same in comparison to a sort by age 
refs #513


nilsthamm (approved):
nice test :ok_hand:

Issue #309, opened 2017-01-11
US_2.4.8: List Views


hendraet:
**As**
pupil

**I want**
uniform list views for displaying data

**In order to**
look at it

**Estimate:** 5

**Acceptance Criteria:**
- [ ] **the application check page will be redesigned independently**
- [ ] text shouldn't collapse if too long (on the left side)
- [ ] add linebreaks when text is too long (size of the container element)
- [ ] "key value pairs" should be displayed next to each other (on application check page)
- [ ] indentation should be the same for all values

nilsthamm:
Changed following list views:
- detail profile (reachable as admin over the role management page)
- request detail (reachable by clicking the title of a request on /requests page) 

Couldn't find more list views that looked strange

hendraet:
:clap:

PR #552, opened 2017-02-07
Edited list views

nilsthamm:
ref #309

### DIFF
--- a/app/assets/stylesheets/profiles.css
+++ b/app/assets/stylesheets/profiles.css
@@ -4,11 +4,11 @@
 */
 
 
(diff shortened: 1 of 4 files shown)


cmfcmf:
@tom95 didn't you also work on these?

hefowe (commented):
@nilsthamm Why do you use a css id for profile_details and a css class for request_details?

nilsthamm:
@hefowe you are right, that makes no sense. I think it was a copy paste error, i will change the profile_details id into a class...

hefowe (approved):
Looks good

Issue #507, opened 2017-02-03
"In Bearbeitung" State of Applications.


tobiasduerschmid:
**As** 
organizer

**I want to**
hide the temporary state (when an organizer pushed the accept / reject button before sending the emails) of an applicant from the user

**in order to**
be able to change it without the applicant noticing the change. 

# Acceptance criteria

- When the event is in the [selection phase](https://github.com/hpi-swt2/workshop-portal/issues/233), the applicant should see "in Bearbeitung" as [state of the application](http://workshop-portal-dev.herokuapp.com/applications)

- Coaches should see the state chosen by organizers (accepted / rejected / alternative)

ekrebs5:
@tobiduer what do coaches see? pending are the "actual" thing (even if actual is temporary and they can't click on it)?

tobiasduerschmid:
@ekrebs5: Coaches should see the state chosen by organizers (accepted / rejected / alternative)

hendraet:
:couplekiss_man_man:

hendraet:
@hefowe Sorry for reopening this but there is bug. If a participant is marked as "angenommen" he sees the warning: please upload letter of agreement

hendraet:
@hefowe nvm this seems to be part of #555

PR #553, opened 2017-02-07
507 in bearbeitung state of applications

hefowe:
Refs #507.


philipp-bode (approved):
Nice tests 👌
You could consider making a factory method for an event in selection phase

Issue #429, opened 2017-01-27
Login Error Rework


hendraet:
**Acceptance Criteria**
- [ ] error on failed login should always say: "E-Mail-Adresse oder Passwort ungültig."

*Reason*: The way it is atm one could find out wihch mails are registered in the system depending on the error message

bjrne:
@hpi-swt2/product-owner-2016 @nstrelow fixed this in [here](https://github.com/hpi-swt2/workshop-portal/blob/dev/config/initializers/devise.rb#L78). Are you sure this is still the case?

hendraet:
@bjrne This isn't what Nils fixed. The problem is that if you type in an incorrect mail address you get "E-Mail-Adresse oder Passwort ungültig.". 
If the mail address is correct it says: "Ungültige Anmeldedaten."
This shouldn't be the case

bjrne:
Ah okay, i thought it was. Alright then :)

nstrelow:
I set the devise paranoid mode. This should make all email queries, so that no one can find out what emails are in the system.

I assume that the string "Ungültige Anmeldedaten" is used by us and is not the default devise string.

So a solution should be to change to string we defined ourselves to the one used for devise.

hendraet:
:raised_hands:

PR #559, opened 2017-02-08
429_login_error_rework

hefowe:
Refs #429.

### DIFF
--- a/config/locales/de.devise.yml
+++ b/config/locales/de.devise.yml
@@ -13,7 +13,7 @@ de:
     failure:
       already_authenticated: "Du bist bereits angemeldet."
       inactive: "Dein Account ist nicht aktiv."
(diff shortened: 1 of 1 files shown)


bjrne:
It should work for now, but I don't really like it. It depends heavily on the correct translation of both strings into every language we might support since we still get two different strings from devise. Might there be another solution?

hefowe:
@bjrne You are right, there should be a better way. Devise itself uses this solution: You can look at the translations https://github.com/tigrish/devise-i18n/tree/master/rails/locales. Devise uses the same string for 'invalid' and 'not_found_in_database' for many languages.

bjrne:
@hefowe Weird. But do you know what the paranoid mode does then?

bjrne (approved):
Since devise itself uses this solution, this should be fine 🗝

Issue #555, opened 2017-02-07
Show "Einverständniserklärung hochladen" after acceptance only


tobiasduerschmid:
**As**
applicant

**I want to**
see the feature for uploading the letter of agreement only when I am accepted

**In order to**
focus on the current task.

# Acceptance criteria

- The text "Bitte lade deine Einverständniserklärung für "Musterworkshop" hoch." with the corresponding file upload is shows after sending the acceptance mails for accepted applicants only.

hendraet:
@Fabian-Sommer please make sure that the user only sees this when he is accepted and "Zusagen/Absagen" are already sent

ekrebs5:
already implemented

Issue #521, opened 2017-02-04
Use I18n for tests


bjrne:
Missing usage in at least:
- [ ] `event_spec.rb`
- [x] `views/application_letters/check.html.erb_spec.rb`

PR #562, opened 2017-02-08
555 show agreement letter upload form only when accepted

Fabian-Sommer:
#555 .
Also refactoring of requires_agreement_letter_for_event (took like 4 minutes).
Also part of #521.


app/models/event.rb
```diff
@@ -402,13 +402,13 @@ def email_addresses_of_participants(all, groups, users)
   # 2. All participants that have to submit an letter of agreement and did do so, ordered by email.
   # 3. All participants that do not have to submit an letter of agreement, ordered by email.
   def compare_participants_by_agreement(participant1, participant2)
-    if participant1.requires_agreement_letter_for_event?(self)
-      if participant2.requires_agreement_letter_for_event?(self)
+    if not participant1.requires_agreement_letter_for_event?(self)
+      if not participant2.requires_agreement_letter_for_event?(self)
```
Blaidd-Drwg: Prefer `unless` to `if not`

app/models/user.rb
```diff
@@ -101,7 +101,7 @@ def agreement_letter_for_event(given_event)
   # @param [Event] given_event
   # @return [Boolean]
   def requires_agreement_letter_for_event?(given_event)
-    return self.older_than_required_age_at_start_date_of_event?(given_event, 18)
+    return (not self.older_than_required_age_at_start_date_of_event?(given_event, 18))
```
Blaidd-Drwg, line 104: `return` is not needed

spec/views/application_letters/check.html.erb_spec.rb
```diff
@@ -2,23 +2,33 @@
...
 RSpec.describe "application_letters/check", type: :view do
 
-  it "should show an upload form for an agreement letter for profiles with an age of <18" do
+  before(:each) do
+    @application_letter = assign(:application_letter, FactoryGirl.create(:application_letter))
+  end
 
+  it "should show an upload form for an agreement letter for profiles with an age of <18 and accepted application" do
     @user = FactoryGirl.create(:user)
+    @user.profile = FactoryGirl.create(:profile, user: @user)
```
Blaidd-Drwg: You can use `@user = FactoryGirl.create(:user_with_profile)` instead

spec/views/application_letters/check.html.erb_spec.rb
```diff
@@ -2,23 +2,33 @@
...
     event = FactoryGirl.create(:event)
-    @application_letter = FactoryGirl.create(:application_letter_accepted, user: @user, event: event)
-    @application_letter.user.profile = FactoryGirl.build(:profile)
+    @application_letter = FactoryGirl.create(:application_letter, status: :accepted, user: @user, event: event)
```
Blaidd-Drwg: What's wrong with application_letter_accepted?

Blaidd-Drwg (changes requested):
Right now you remove the button in the front end. Somebody who knows the url and has too much time on their hands could still make a POST request and upload an AgreementLetter. Could you also modify the `AgreementLettersController` so that it's impossible to upload an AgreementLetter if you don't need to?

spec/views/application_letters/check.html.erb_spec.rb
Fabian-Sommer: Defining a new factory to set a single attribute isn't good style, since you can only use one factory. So I think application_letter_accepted should never have existed in the first place, since attributes can easily be set using this way. Both work for now. So what is wrong with this?

app/models/user.rb
Fabian-Sommer: While not needed, I think it makes the code easier to understand

Fabian-Sommer:
@Blaidd-Drwg I think the odds of the POST-request vulnerability ever being used by anyone are extremely close to 0. I think if it does happen anyway I am convinced it would in no way cause any harm to anyone. The US is phrased like "upload field cannot be seen". In addition, the "vulnerability" existed before and was partly implemented by you. Feel free to open a new issue or to fix it yourself.

app/views/events/participants.html.erb
```diff
@@ -41,7 +41,7 @@
           <%= participant.email %>
         </td>
         <td>
-          <% if participant.requires_agreement_letter_for_event?(@event) %>
+          <% if not participant.requires_agreement_letter_for_event?(@event) %>
```
Blaidd-Drwg, line 44: `unless`

app/models/user.rb
Blaidd-Drwg: IMO it does not, since when developing in Ruby you are used to the last statement in the function being the return value. In addition the [style guide](https://github.com/bbatsov/ruby-style-guide#no-explicit-return) defines it as a good practice. But I'll leave the final choice to you, I'm going to approve either way :wink:

Blaidd-Drwg:
>In addition, the "vulnerability" existed before and was partly implemented by you.

I am fully aware of that, though I don't see why this would be relevant. In any case, you have a point, the US does not require it. I'll inform the POs on the issue page and let them decide if they want to open a new issue for that.

Blaidd-Drwg (approved):

PR #568, opened 2017-02-08
Presentation data

frederike-ramin:
Create sensible data for use in the final presentation.


timbasel (approved):
Looks good to me

querenker (approved):
I fixed some typos in the rejection mail template. Now it should be fine :)

db/sample_data/events.rb
```diff
@@ -74,7 +74,7 @@ def event_gongakrobatik
       end_date: Date.new(2021, 03, 05)
   )
   Event.new(
-      name: 'Einführung in die Kunst der Gongakrobatik',
+      name: 'Gongakrobatik',
```
bjrne, line 77: Why did you do this? Only for the presentation because long titles don't work well at the moment? If this is the case, please open a bug/technical debt/enhancement issue so we don't forget this (Since titles of that length shouldn't be a problem in the future 🙂 )

app/assets/stylesheets/events.css
```diff
@@ -106,3 +106,7 @@
     height: 150px;
     margin-bottom: 15px;
 }
+
+.text-success{
+    color: #000000;
+}
```
cmfcmf: This doesn't look right. Success text should most likely be green.

db/sample_data/email_templates.rb
```diff
@@ -0,0 +1,60 @@
+def email_template_acceptance
+  EmailTemplate.new(
+    status: :acceptance,
+    hide_recipients: false,
+    subject: "Zusage für TODO",
```
cmfcmf, line 5: TODO?

tobiasduerschmid: TODO schould be changed by the organizer (enter camp name, etc.) This is intended :)

db/sample_data/events.rb
tobiasduerschmid: Just because we do not want to have so huge names in order to keep the UI simple.

db/sample_data/email_templates.rb
cmfcmf: Oh, right. Got it!

tobiasduerschmid (approved):
Now the texts should be great :)

db/sample_data/events.rb
bjrne: So it is more or less just a presentation thing? Future users might still choose a long name...

app/assets/stylesheets/events.css
```diff
@@ -118,3 +118,7 @@
     height: 150px;
     margin-bottom: 15px;
 }
+
+.text-success{
```
Baschdl: Why did you change a css file?

bjrne (dismissed):
Formal block based on @Baschdl's question

app/assets/stylesheets/events.css
tobiasduerschmid: Because the text color was white and nobody could read it.

db/sample_data/events.rb
tobiasduerschmid: Yes

bjrne (approved):
🍭

PR #573, opened 2017-02-08
Attach agreement letter file to acceptance mails

tom95:


aloeser (approved):
Worked for me locally.

app/models/email.rb
```diff
@@ -41,6 +41,13 @@ def persisted?
 
   private
 
+  def get_agreement_letter_attachment
+    {
+      name: (I18n.t 'emails.agreement_letter_attachment'),
+      content: File.read(Rails.root.join('storage', 'agreement_letters', 'real_agreement_letter.pdf'))
```
aloeser: I'd suggest to read the path from the config, either in this PR, or later (we need a technical debt issue then)

config/application.rb
```diff
@@ -31,5 +31,7 @@ class ApplicationLetter < Rails::Application
 
     config.from_address = ENV['SMTP_ADDRESS'] ? ENV['SMTP_ADDRESS'] : 'noreply@localhost'
     config.reply_to_address = config.from_address
+
+    config.empty_agreement_letter_path = File.read(Rails.root.join('storage', 'agreement_letters', 'real_agreement_letter.pdf'))
```
cmfcmf: I think you should remove the path suffix, because it Really is the content, not the path.

tom95: @cmfcmf fun fact: I meant to only put the path :P I'll change it so that it won't be kept in memory

cmfcmf:
Even better!

Tom Beckmann <notifications@github.com> schrieb am Do., 9. Feb. 2017, 08:43:

Issue #572, opened 2017-02-08
Redesign of startpage


hendraet:
**Acceptance Criteria**
- [ ] startpage should look like this:
![startseite_hpi_hintergrund](https://cloud.githubusercontent.com/assets/16731894/22758563/7ae934b4-ee4f-11e6-832d-3b7f0ef2aa7f.png)
- [ ] bg image: https://hpi.de/fileadmin/user_upload/hpi/bilder/teaser_channels/architektur_campus/hpi_main_building_keyvisual_1020x420.jpg
- keep yellow color for the other boxes

nstrelow:
@hendraet could you please provide an image with a higher resolution

This one works, but on a FullHD screen it doesn't look great

MasterCarl:
@nstrelow Here's a higher resolution image: <https://hpi.de/fileadmin/user_upload/hpi/navigation/25_presse/25_bildmaterial/05_Fotos/10_campus/HPI_Hauptgebaeude_Tag_3054x2036.jpg>

PR #587, opened 2017-02-09
add background to top of the welcome page

nstrelow:
refs #572

The background image should probably be replaced by one with a higher resolution.


app/assets/stylesheets/application.css
```diff
@@ -29,6 +29,17 @@ body, footer, main, .front-section {
...
 
+.banner {
+    position: absolute;
+    top: 50px;
+    left: 0px;
+    right: 0px;
+    width: auto;
+    height: 100%;
+    z-index: -1;
+    filter: brightness(50%);
```
tom95: In my opinion you should apply the filter directly to the image data instead of leaving it to the browser, since this is the only place where we need it, so we might as well make sure the text doesn't become completely illegible on non-supported browsers.

Additionally, the current layouting code causes overflow on my device (1265px width) and at the bottom, the orange shows through. Working with height 100% here doesn't make much sense unless you having a surronding relatively positioned element that is exactly the height that you want. Right now, the image will scale when I vertically resize my browser.
The best way to go would probably be a `background-image` with `background-size: cover` and then make sure you kill all the marging around the topmost section so that no orange shows through.

app/views/layouts/application.html.erb
```diff
@@ -32,6 +32,8 @@
   <% end %>
 <% end %>
 
+<%= image_tag("hpi_main_building_keyvisual_1020x420.jpg", :alt => "HPI Background", :class => 'banner') %>
```
tom95: layouts/application means on any page there is :P you probably meant to put this in views/application/index

tom95 (dismissed):

app/views/layouts/application.html.erb
radscheit: This image is way to small, it should have at least 1920 px in width in order to fit for various screen resolutions.

radscheit: When you the `alt` attribute, you should make it translatable.

app/assets/stylesheets/custom_bootstrap/custom_bootstrap.less
```diff
@@ -168,6 +168,10 @@ a:hover h1, a:hover h2, a:hover h3, a:hover h4 {
 #wsp-welcome {
 	padding-top: 0px !important;
 
+    h1, h4, p {
```
radscheit: Wrong Indention and for consitency we use Hexcodes for colors.

radscheit (changes requested):

app/assets/stylesheets/application.css
nstrelow: thanks for the great suggestions 😄

tom95:
The branch has been updated for a new review now

app/assets/stylesheets/custom_bootstrap/custom_bootstrap.less
```diff
@@ -184,6 +184,10 @@ a:hover h1, a:hover h2, a:hover h3, a:hover h4 {
 	
 	padding-top: 0px !important;
 
+    h1, h4, p {
```
radscheit, line 187: Still wrong indentions.

app/views/application/index.html.erb
```diff
@@ -1,5 +1,6 @@
-<section class="front-section">
+<section class="front-section banner" style="background-image: url('<%= image_path('hpi_main_building_keyvisual_1020x420.jpg') %>')">
```
radscheit, line 1: Still a too small image.

radscheit (changes requested):

app/views/application/index.html.erb
radscheit: And why do you use inline css?

tom95: need to get the image_path, can't do that in the CSS/less. we won't get a larger version of the image until tomorrow

radscheit (approved):

PR #589, opened 2017-02-09
60 5.3a sort materials

Blaidd-Drwg:
#60


cmfcmf:
Let me look at this before merging, please.

app/controllers/events_controller.rb
```diff
@@ -283,20 +250,62 @@ def filter_application_letters(application_letters)
...
     end
 
-    # Gets all file names stored in the material storage of the event
-    #
-    # @param [Event]
-    # @return [Array of Strings]
-    def get_material_files(event)
-      material_path = event.material_path
-      File.exists?(material_path) ? Dir.glob(File.join(material_path, "*")) : []
+    def build_files(material_path, current_path)
```
cmfcmf, line 262: Consider using `Find` to simplify this: http://stackoverflow.com/a/15698165/2560557

app/controllers/materials_controller.rb
```diff
@@ -0,0 +1,221 @@
...
+
+    if !is_file?(file)
+      redirect_to event_path(@event), alert: t("events.material_area.no_file_given") and return
+    elsif invalid_pathname?(path) || invalid_filename?(file.original_filename)
+      redirect_to event_path(@event), alert: t("events.material_area.invalid_path_given") and return
+    elsif !File.directory?(File.join(@event.material_path, path))
+      redirect_to event_path(@event), alert: t("events.material_area.download_file_not_found") and return
+    end
+
+    file_path = File.join(@event.material_path, path, file.original_filename)
```
cmfcmf, line 23: Since the `original_filename` is user controlled, I am not 100% sure it can't contain `../`. Something to verify..

app/controllers/materials_controller.rb
```diff
@@ -0,0 +1,221 @@
...
+    end
+
+    redirect_to event_path(@event), notice: I18n.t("events.material_area.success_message")
+  end
+
+  # POST /events/1/download_material
+  def download_material
+    file_full_path = File.join(@event.material_path, params[:file].to_s)
+
+    if !params.has_key?(:file)
```
cmfcmf, line 37: Shouldn't this check happen *before* using `params[:file]` above?

app/controllers/materials_controller.rb
```diff
@@ -0,0 +1,221 @@
...
+    rescue IOError
+      redirect_to event_path(@event), alert: t("events.material_area.saving_fails") and return
+    end
+
+    redirect_to event_path(@event), notice: I18n.t("events.material_area.success_message")
+  end
+
+  # POST /events/1/download_material
+  def download_material
+    file_full_path = File.join(@event.material_path, params[:file].to_s)
```
cmfcmf, line 35: Doesn't this allow using `../../..` as `file` parameter?

app/controllers/materials_controller.rb
```diff
@@ -0,0 +1,221 @@
...
+    if File.directory?(file_full_path)
+      send_zipped_materials(file_full_path)
+    else
+      send_file file_full_path, x_sendfile: true
+    end
+  end
+
+  def move_material
+    from_path = File.join(@event.material_path, params[:from].to_s)
+    to_path = File.join(@event.material_path, params[:to].to_s)
```
cmfcmf, line 54: Same here: Shouldn't `:to` and `:from` be checked before using them. Also, can't they be used to move the file anywhere?

cmfcmf: This applies for all other methods below as well.

app/controllers/materials_controller.rb
```diff
@@ -0,0 +1,221 @@
+require 'rubygems'
```
cmfcmf, line 1: This isn't a *thin controller* at all.

cmfcmf (changes requested):
I haven't gone through the whole PR, but tried to find some security loopholes.

app/controllers/materials_controller.rb
Blaidd-Drwg: Most of it consists in checking if the parameters are valid, so we figured that's a task for the controller. Open to suggestions, though, what parts do you think should be moved to another class?

app/controllers/materials_controller.rb
Blaidd-Drwg: `invalid_filename?(file.original_filename)` in line 17 checks for both `/` and `..` in the filename

app/controllers/materials_controller.rb
Blaidd-Drwg: I moved the assignment to `file_full_path` up so that I could write the `if..elsif..` construct to check for errors without interruptions, I thought it made the code clearer. I agree that it looks kind of weird to assign the value before checking whether it's valid, but since `find_full_path` is only used after this check, it doesn't change anything to the way the method works.

app/controllers/materials_controller.rb
Blaidd-Drwg: We always use `invalid_pathname?` (line 211) to validate the given paths. If the pathnames start with `/` or contain `..`, we return and display an error.

app/controllers/events_controller.rb
Blaidd-Drwg: Thanks, I'll look into it :slightly_smiling_face:

app/controllers/materials_controller.rb
cmfcmf: I see 👍

app/controllers/materials_controller.rb
cmfcmf: I think most of the private methods could be moved into some sort of FileHandler class.

Raandom: Also, making a Fat Model when you don't have any model you work on is rather hard. Dont know wheather it's sensefull to move them to a helper/api-class or something like that, as the actual move/create and so on methods are usually rather short or even one liners. Its mainly validation making it big/unhandy.

app/controllers/materials_controller.rb
cmfcmf: Let's hope this never runs on Windows :P No, it's fine, I guess.

app/controllers/materials_controller.rb
cmfcmf: I had assumed `params[:file]` would error if `!params.has_key?(:file)`. But if it doesn't, fine!

app/controllers/materials_controller.rb
cmfcmf: No one prevents you from creating a service: https://blog.engineyard.com/2014/keeping-your-rails-controllers-dry-with-services

app/controllers/materials_controller.rb
Raandom: Nah. We are using Pathname.absolute? for validation and File::SEPERATOR for checks. Should run fine on both systems.

cmfcmf: > Nah. We are using Pathname.absolute? for validation and File::SEPERATOR for checks. Should run fine on both systems.

Event better 👍

bjrne:
@hpi-schuelerklub/swt2 I'll close it for now. My reasons are:
* Security considerations
* Merge conflicts
* Huge PR
* Probably not a priority right now
* There is probably a gem for file-handling that supports this

We might look again at this when looking at #60.

PR #613, opened 2017-02-09
Description out of applications pdf

radscheit:


lib/pdf_generation/applications_pdf.rb
```diff
@@ -8,6 +8,7 @@ class ApplicationsPDF
   # @return [String] the generated PDF
   def self.generate(event)
     self.new(event).create.render
+
```
frederike-ramin: Don't think thats necessary.

frederike-ramin (changes requested):

lib/pdf_generation/applications_pdf.rb
radscheit: Ist schon weg.

frederike-ramin (approved):

PR #622, opened 2017-02-10
58 3.16 personal email refactoring

timbasel:
Refactoring of the email system-

- Moved Attachment creation out of  the email model
- Unified the status variable in Email


aloeser (changes requested):
I dont think we want to go back to accepted and rejected for email template states.
We renamed them from accepted to acceptance in an earlier PR, in order to avoid mixing two things that are equal by accident, but dont mean the same. Also, you rather send an acceptance than accepted mail.

And I'm not sure if I would rename get_email_template_status to get_status, the first one seems more precise to me

app/models/agreement_letter.rb
```diff
@@ -103,4 +103,11 @@ def unable_to_open? file
       true
     end
   end
+
+  def self.get_attachment
```
ekrebs5: I would leave this as get agreement letter attachment, since it is more precise and other attachments in emails are very likely

ekrebs5 (approved):

aloeser (approved):

PR #634, opened 2017-05-10
refactor application letter factory

aloeser:
we now use traits instead of inheritance


aloeser:
One thing I'm not sure about is whether to put the traits into or below the factories, like

```
factory :application_letter do

    trait :accepted do
          status :acccepted
    end
end

trait :rejected do
    status :rejected
end
```

Any comments or ideas?

aloeser:
Nevermind, 1min debugging with `rails console` helped. Putting them below factories makes them globally available, so that you can run `FactoryGirl.build(:event, :accepted)` which doesnt make sense.
I'll change it soon

spec/controllers/email_controller_spec.rb
```diff
@@ -15,7 +15,7 @@
 
     context "with valid accepted applications" do
       before :each do
-        @application = FactoryGirl.create(:application_letter_accepted, event: @event, user: FactoryGirl.build(:user))
+        @application = FactoryGirl.create(:application_letter, :accepted, event: @event, user: FactoryGirl.build(:user))
```
querenker: Is the user association in this create call even necessary? I think [this association definition](https://github.com/hpi-schuelerklub/workshop-portal/blob/35d5b8dd50ac72141dd33b35200a3682b5712588/spec/factories/application_letters.rb#L22) should be enough?

spec/controllers/email_controller_spec.rb
```diff
@@ -33,7 +33,7 @@
 
     context "with valid rejected applications" do
       before :each do
-        @application = FactoryGirl.create(:application_letter_rejected, event: @event, user: FactoryGirl.build(:user))
+        @application = FactoryGirl.create(:application_letter, :rejected, event: @event, user: FactoryGirl.build(:user))
```
querenker: User association again

querenker (changes requested):
The traits are very nice 👍

querenker (approved):

bjrne (approved):
Good 🏃‍♀️

Issue #603, opened 2017-02-09
Valid HPI email addresses in sample data


chrisma:
The sample data (`db/sample_data/users.rb`) currently contains the valid email address `admin@hpi.de`

![image](https://cloud.githubusercontent.com/assets/1652117/22790566/e5cdf93a-eee6-11e6-9674-008a7f0a9350.png)

annkatrinkuessner:
additionally replace the valid e-mail addresses of Tobias and Hendrik.

PR #635, opened 2017-05-11
#603 remove valid email addresses from sample data

corinnaj:
We should probably also think about removing german words from the sample email addresses entirely.
I also changed the admin email address in the Wiki.


aloeser (approved):
did a `grep "hpi.de" -r app db`., found no valid addresses

PR #648, opened 2017-05-22
Correctly annotate email test with js

bjrne:
If anyone knows why this test needs JS, please tell me 😁


querenker (approved):
I can't see any reason for including js either.

cmfcmf:
I'll check this as well.

PR #649, opened 2017-05-22
505 past date validations

### LINKED ISSUE #505: Remove past date validations
After discussion with @tom95 we decided we no longer need these validations since it is unnecessary code overhead and prevents past events to be edited.

bjrne:
Closes #505.

### DIFF
--- a/spec/features/event_spec.rb
+++ b/spec/features/event_spec.rb
@@ -16,7 +16,7 @@ describe "Event", type: :feature do
 
     it "should not list past events" do
       current_event = FactoryGirl.create :event
-      past_event = FactoryGirl.create :event, :in_the_past_valid
+      past_event = FactoryGirl.create :event, :in_the_past
 
       visit events_path
       expect(page).to have_text(current_event.name)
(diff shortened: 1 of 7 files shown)


ekrebs5 (approved):

spec/features/event_spec.rb
```diff
@@ -135,12 +135,12 @@
       end
     end
 
-    it "should not allow dates in the past" do
+    it "should allow dates in the past" do
```
querenker: I am not sure if this test communicates the right thing. If I were a new developer to this project I would ask what's the purpose of creating events in the past. Allowing old events to edit makes sense, but creating "old" events?

querenker (changes requested):
I think this one test is worth a discussion, but otherwise it looks good to me 👍

spec/features/event_spec.rb
bjrne: Wow, didn't think of that. I like the idea, would changing the test action to editing an existing event and saving it without errors be what you suggested? (Did that by now, please look at the new commits 😃 )

querenker (approved):

spec/features/event_spec.rb
querenker: Looks good to me 👍

PR #650, opened 2017-05-22
510 red stars event

### LINKED ISSUE #510: Event creation form does not have the red stars for required fields
It is unclear which fields are required.

@hpi-swt2/product-owner-2016

bjrne:
Closes #510.

While I was at it I corrected some style and convention 'errors'. If you want, I can move them to a separate PR where we could do this once an for all in all files as well. Or we do this on the fly with everything we touch 🎵

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -28,6 +28,8 @@ class Event < ActiveRecord::Base
   has_many :participant_groups
   has_many :date_ranges
   accepts_nested_attributes_for :date_ranges
+  validates_presence_of :name
+  validates_presence_of :description
   validates :max_participants, numericality: { only_integer: true, greater_than: 0 }
   validate :has_date_ranges
   validates_presence_of :application_deadline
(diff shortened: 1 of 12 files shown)


ekrebs5 (approved):

querenker (approved):
Looks good to me 👍

app/models/event.rb
```diff
@@ -28,6 +28,8 @@ class Event < ActiveRecord::Base
   has_many :participant_groups
   has_many :date_ranges
   accepts_nested_attributes_for :date_ranges
+  validates_presence_of :name
+  validates_presence_of :description
```
corinnaj: This might be nitpicking, but in other models we put all "validates_presence_of" in one line, separated by commas.

bjrne: Nah, I like it, This way I got to refator all `validate :xyz, presence: true` to this more readable form (at least in my eyes).

PR #665, opened 2017-06-25
fixes #520 misleading draft actions

### LINKED ISSUE #520: Draft actions are misleading on event creation page
When using the indicative form of verbs on buttons, user might expect something to happen. I'd prefer a more passive version. Moreover, when finally clicking on the create button on the bottom, it is not obvious what will happen since it says "Erstellen" independently of my choice. @hpi-swt2/product-owner-2016 

bildschirmfoto 2017-02-04 um 22 45 32

corinnaj:
We now have actual buttons, instead of radio buttons for drafts. It is not possible to unpublish events, I wasn't sure if that is a useful feature, but it should be possible to reintroduce that quickly, if necessary.

I wasn't sure what our conventions for buttons in tests are, so they might have to be refactored.


spec/features/event_image_spec.rb
```diff
@@ -19,20 +18,20 @@
...
 
     it "should not change the image if validation for a new image fails" do
-      click_button I18n.t(".events.form.create")
+      click_button "create"
       previous_image = Event.last.image
 
       visit edit_event_path(Event.last)
       attach_file "event[custom_image]", File.join(Rails.root + 'spec/testfiles/too_small_image.png')
-      click_button I18n.t('.events.form.update')
+      click_button "update"
```
bjrne: Why did you remove the I18n here?

bjrne (approved):
I don't know the motivation for removing all the internationalization lines, but if it's legitimate, then this is ready to go 🏖

spec/features/event_image_spec.rb
corinnaj: That's what I meant about button conventions in tests. You can use either the button label to click the button, which is the I18n line, or the button id, which is "update".
We mostly use the labels currently, but I used the ids in some places, because the labels got quite long.

bjrne: Okay. Since this never lets a line exceed 80 chars or even create a new line, I'm in favor of the I18n version, but I'd like more opinions on this here @hpi-schuelerklub/swt2

bjrne (changes requested):
Attention, there are git tags in the latest commit

corinnaj:
I am not sure what you mean... I removed the one remaining git conflict, I think it should be fine now.

spec/features/event_image_spec.rb
```diff
@@ -9,14 +9,9 @@
     before :each do
       login_as(FactoryGirl.create(:user, role: :organizer), :scope => :user)
       visit new_event_path
-<<<<<<< HEAD
```
bjrne: Ah, I thought this was green! sorry :D

bjrne (approved):

annkatrinkuessner:
We decided that our style guideline is to use internationalization for the tests instead of ids. It would be amazing if this could be adapted before merging :)

corinnaj:
I updated all test to use I18n instead of ids. There is only one rubocop issue remaining, which I propose to fix in another PR, as it's more of a general problem and not part of this issue.

querenker:
That the classes are too long is indeed a general problem... The rubocop todo file is now adjusted accordingly :)

PR #674, opened 2017-08-09
Openid email fix

### LINKED ISSUE #664: Error message for openID without email

bjrne:
Closes #664.


spec/features/user_hpiopenid_spec.rb
```diff
@@ -34,12 +34,42 @@
...
+                                                                       }
+                                                                   })
+    visit new_user_registration_path
+    click_link I18n.t('devise.shared.links.sign_in_with_provider', :provider => 'HPI OpenID')
+
+    expect(page).to have_css(".alert-danger")
+    expect(page).to have_text (I18n.t('users.openid.missing_email'))
+  end
+
+  scenario 'Account creation fails due to missing email' do
```
querenker: I am not sure how we should handle code duplication in tests. Currently the both tests are very similar.

querenker (changes requested):
I am not sure about the code duplication in the tests, but otherwise it looks good to me :)

spec/features/user_hpiopenid_spec.rb
bjrne: They are, but I don't know a better way. Would you create a method for the part that is the same? I think I'd find it weird to have the `expect` in a function and not the test itself. Unfortunately the big part in the beginning is the one that differs from test to test...

querenker: I am not sure if these will make the code better, but I have some suggestions:

1. Loop around the scenarios:
``` ruby
['', nil].each {|invalid_email|
  scenario "Account creation fails if email equals #{invalid_email}" do
    OmniAuth.config.mock_auth[:hpiopenid] = OmniAuth::AuthHash.new({
                                                                       :provider => @provider,
                                                                       :uid => @uid,
                                                                       :info => {
                                                                           :email => invalid_email
                                                                       }
                                                                   })
    visit new_user_registration_path
    click_link I18n.t('devise.shared.links.sign_in_with_provider', :provider => 'HPI OpenID')

    expect(page).to have_css(".alert-danger")
    expect(page).to have_text (I18n.t('users.openid.missing_email'))
  end
}
```

2. Combine both into one scenario and add the loop there

3. Maybe parts of the scenario could be extracted into seperate methods, such as to create a valid AuthHash and change the value in the test to make them invalid?

querenker (approved):

spec/features/user_hpiopenid_spec.rb
querenker: I have withdrawn my change request, because this is not a problem directly related to this PR, but a discussion in general how test should be lool like :)

bjrne: I really like the loop idea since this also extracts the nil values to a more visible place and makes the whole test more readable, not to mention it also solves the duplication problem here. I will implement your idea soon.

PR #676, opened 2017-08-09
Fixes #636 add coding style guide

### LINKED ISSUE #636: Add coding style guide
I think [this community driven coding style guide](https://github.com/bbatsov/ruby-style-guide) looks promising. There is also good tool support for example [to comment automatically on style violations in GitHub pull requests](https://houndci.com/). Is it worth a try?

querenker:
This PR introduces a coding standard, which can be automatically checked with [rubocop](rubocop.readthedocs.io)

The `.rubocop_todo.rb` file lists all the current problem. By running the `rubocop` command from the command line only the new introduced problems will be displayed. If a problem listed in the `.rubocop_todo.rb` should be fixed it can be enabled and the rubocop command will list the problems.


app/models/agreement_letter.rb
```diff
@@ -65,49 +65,48 @@ def set_path
...
+    elsif too_big?(file)
+      errors.add(I18n.t('agreement_letters.file_error'), I18n.t('agreement_letters.file_too_big'))
+      false
+    elsif wrong_filetype?(file)
+      errors.add(I18n.t('agreement_letters.file_error'), I18n.t('agreement_letters.wrong_filetype'))
+      false
+    elsif unable_to_open?(file)
+      errors.add(I18n.t('agreement_letters.file_error'), I18n.t('agreement_letters.corrupt_document'))
       false
-    rescue PDF::Reader::UnsupportedFeatureError, PDF::Reader::MalformedPDFError
```
bjrne, line 107: Is it okay to delete this line?

app/models/agreement_letter.rb
```diff
@@ -15,7 +15,7 @@ class AgreementLetter < ActiveRecord::Base
   belongs_to :event
   validates_presence_of :user, :event, :path
   MAX_SIZE = 300_000_000
-  ALLOWED_MIMETYPE = "application/pdf"
+  ALLOWED_MIMETYPE = 'application/pdf'.freeze
```
bjrne, line 23: What does freeze do?

bjrne (approved):
Wow, I like it a lot. I suggest we merge this ASAP before this gets difficult to keep conflict free 🚥

app/models/agreement_letter.rb
querenker: [`freeze`](https://apidock.com/ruby/Object/freeze) prevents further modification of the object. A RuntimeError will be raised if modification is attempted.

app/models/agreement_letter.rb
querenker: The presentation of this diff is a little bit weired, but I think this line is now [here](https://github.com/hpi-schuelerklub/workshop-portal/pull/676/files/02458dece54bc25584346eccc9e96c1895f77d93#diff-b886d6b9c92f8c631ec0fad6268523ebR102)

PR #681, opened 2017-08-09
672 roles are no longer in a hierarchy

aloeser:
all roles are separate, abilities are not inherited between roles
the user has exactly one role


annkatrinkuessner (approved):
looks good

app/models/user.rb
```diff
@@ -38,7 +38,7 @@ class User < ActiveRecord::Base
   def role?(base_role)
     return false unless role
 
-    ROLES.index(base_role) <= ROLES.index(role.to_sym)
+    ROLES.index(base_role) == ROLES.index(role.to_sym)
```
cmfcmf: This looks like it could be simplified to `base_role.to_sym == role.to_sym`. The `ROLES` variable could either be completely removed or instead used to verify that base_role is valid `[PSEUDOCODE] unless base_role is part of ROLES throw some error`.

bjrne:
@aloeser You could maybe create an issue for the removal of the `manage` ability and link it here :)

Update: see #706

app/models/user.rb
aloeser: How do we want to deal with invalid roles? It is a getter function, so we cannot get into an invalid state with an invalid role given. With an invalid role it will always return false (which is correct), so we dont actually have a "real" error.
However, throwing an error might make it easier to find bugs

bjrne: I like the explicit (error) version I think. Shouldn't happen too much and the log would be waaaay more helpful then.

app/models/ability.rb
```diff
@@ -4,7 +4,7 @@ class Ability
   def initialize(user)
     # Define abilities for the passed in user here. For example:
     #
-    #   user ||= User.new # guest user (not logged in)
+    #   user ||= User.new # guest user [not logged in]
```
cmfcmf: A bit eager with the brace removal :-D

app/models/ability.rb
```diff
@@ -29,57 +29,56 @@ def initialize(user)
...
-      can %i(manage set_contact_person set_notes), Request
-      cannot :apply, Event
-      cannot :view_apply_button, Event
-      can %i(edit update destroy), Event
-      can [:update], ParticipantGroup
+      can :update, ParticipantGroup
 
       # Organizers can update user roles of pupil, coach and organizer, but cannot manage admins and cannot update a role to admin
-      can :manage, User, role: %w(pupil coach organizer)
+      can :manage, User, role: %w[pupil coach organizer]
```
cmfcmf: Why is this `%w`?

cmfcmf (changes requested):
There are some minor things I found - mostly formatting. I'll now look at the permission changes themselves.

cmfcmf (approved):
Noice!

PR #682, opened 2017-08-09
remove valid last names from sample data

annkatrinkuessner:


bjrne (approved):
A little uncreative, but works. 👨‍⚕️

querenker (approved):
It is indeed a little bit uncreative, but looks good to me 👍

PR #690, opened 2017-08-23
Update badge links and use SVG heroku badges

querenker:
The badges should now be correctly linked with this repo.

But some challenges/questions are pending:
* The coverage badge for the production branch is currently not available, maybe a setting in coveralls is missing? 
* Is the Errbit link up to date? Does someone has the login credentials and is it in use? And should we link it there?

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,8 @@ A Ruby on Rails app to manage workshops
 
 Branch | Travis CI  | Code Analysis | Heroku Deploy | Errbit
 ------ | ---------- | ------------- | ------------- | ------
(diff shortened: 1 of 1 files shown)


bjrne:
We currently have no production branch anymore, our dev branch is the standard one and the one we deploy to heroku. Thus, the link cannot work at the moment.
EDIT: Weird, we still have it but coveralls is not able to detect it 🙁 

I think Errbit will be a separate issue regarding the final deployment and monitoring solutions, but I'd look into this when we get the final deployment inside of HPI working.

bjrne (approved):
I don't know why coveralls is so weird with the production branch. If you want, you can merge this now so at least have the working version of the dev badge. I think Errbit is another issue

PR #712, opened 2018-10-26
move all login method to macros

lHydra:
[Issue 684](https://github.com/hpi-schuelerklub/workshop-portal/issues/684)


bjrne (approved):
Looks good to me. Thanks for your contribution! We require a second review, as soon as it is in, we'll merge this.

